PR #8, opened 2021-04-18
Support missing expects method in MigrateAtToConsecutiveExpectations

### LINKED ISSUE #7: Support missing expects method while using MigrateAtToConsecutiveExpectations
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | v0.10.6              |
| Installed as   | prefixed Rector                           |

Using the `MigrateAtToConsecutiveExpectationsRector` rule. If a mock was configured without calling the `expects()` method, an error would be threw.

Omitting `expects()` while configuring a mock is valid syntax.

# Minimal PHP Code Causing Issue
```php
<?php

namespace Rector\PHPUnit\Tests\Rector\ClassMethod\MigrateAtToConsecutiveExpectationsRector\Fixture;

class Foo
{
    public function someMethod()
    {
    }
}
final class SkipMissingExpects extends \PHPUnit\Framework\TestCase
{
    protected $mock;

    protected function setUp(): void
    {
        parent::setUp();

        $this->mock = $this->createMock(Foo::class);
    }

    public function test(): void
    {
        $this->mock
            ->method('someMethod')
            ->willReturn(1);

        $mock = $this->createMock(Foo::class);

        $mock
            ->method('someOtherMethod')
            ->willReturn(1);
    }
}
```

```php
return static function (ContainerConfigurator $containerConfigurator): void {

    $services->set(\Rector\PHPUnit\Rector\ClassMethod\MigrateAtToConsecutiveExpectationsRector::class);
};
```

# Actual Behaviour
```
 [ERROR] Could not process "..." file,
         due to:                                                                                                        
         "Argument 1 passed to Rector\PHPUnit\NodeAnalyzer\ExpectationAnalyzer::isValidExpectsCall() must be an instance
         of PhpParser\Node\Expr\MethodCall, instance of PhpParser\Node\Expr\PropertyFetch given, called in              
         vendor/rector/rector-prefixed/vendor/rector/rector-phpunit/src/NodeAnalyzer/ExpectationAnalyzer.php:49".
```

JoshuaLicense:
Closes #7.

Wasn't 100% sure of the best way to handle this - Happy to make changes to the fix based on feedback.

### DIFF
--- a/tests/Rector/ClassMethod/MigrateAtToConsecutiveExpectationsRector/Fixture/skip_missing_expects.php.inc
+++ b/tests/Rector/ClassMethod/MigrateAtToConsecutiveExpectationsRector/Fixture/skip_missing_expects.php.inc
@@ -10,58 +10,8 @@ class Foo
 }
 final class SkipMissingExpects extends \PHPUnit\Framework\TestCase
 {
-    protected $mock;
-
-    protected function setUp(): void
-    {
-        parent::setUp();
-
-        $this->mock = $this->createMock(Foo::class);
-    }
-
     public function test(): void
     {
-        $this->mock
-            ->method('someMethod')
-            ->willReturn(1);
-
-        $mock = $this->createMock(Foo::class);
-
-        $mock
-            ->method('someOtherMethod')
-            ->willReturn(1);
-    }
-}
-
-?>
------
-<?php
-
-namespace Rector\PHPUnit\Tests\Rector\ClassMethod\MigrateAtToConsecutiveExpectationsRector\Fixture;
-
-class Foo
-{
-    public function someMethod()
-    {
-    }
-}
-final class SkipMissingExpects extends \PHPUnit\Framework\TestCase
-{
-    protected $mock;
-
-    protected function setUp(): void
-    {
-        parent::setUp();
-
-        $this->mock = $this->createMock(Foo::class);
-    }
-
-    public function test(): void
-    {
-        $this->mock
-            ->method('someMethod')
-            ->willReturn(1);
-
         $mock = $this->createMock(Foo::class);
 
         $mock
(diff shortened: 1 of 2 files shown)


tests/Rector/ClassMethod/MigrateAtToConsecutiveExpectationsRector/Fixture/skip_missing_expects.php.inc
```diff
@@ -0,0 +1,73 @@
...
+
+        $mock = $this->createMock(Foo::class);
+
+        $mock
+            ->method('someOtherMethod')
+            ->willReturn(1);
+    }
+}
+
+?>
```
TomasVotruba, line 73: Hi, could you simplify test to fewer lines? The diff of changes should be more obvious

JoshuaLicense: Sure - how's it look now?

I've removed the comparison as it should be skipped (and not refactored).

TomasVotruba: Better, thanks :+1: 

What happens with this code without the change?

JoshuaLicense: It will throw an error.

If the `expects()` is missing, the `var` property of `$methodCall` (line 151) will be a `PropertyFetch` or `Variable` (depending where the object is stored) rather than the expected `MethodCall` object (for `expects()`).

Error would be threw at line 67 - `isValidExpectsCall()` which expects a `MethodCall` object as a parameter.

TomasVotruba: I see, thanks :+1:

PR #10, opened 2021-04-18
Handle mocks stored as properties in MigrateAtToConsecutiveExpectation

### LINKED ISSUE #9: Support mocks stored as properties while using MigrateAtToConsecutiveExpectations
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | v0.10.6              |
| Installed as   | prefixed Rector                           |

Using the `MigrateAtToConsecutiveExpectationsRector` rule. If a mock was configured in the `setUp()` method and stored as a property, the `at()` would be ignore and not refactored.

# Minimal PHP Code Causing Issue
```php
class Foo
{
    public function someMethod()
    {
    }
}
final class HandlePropertyFetch extends \PHPUnit\Framework\TestCase
{
    protected $mock;

    protected function setUp(): void
    {
        parent::setUp();

        $this->mock = $this->createMock(Foo::class);
    }

    public function test(): void
    {
        $this->mock
            ->expects($this->at(0))
            ->method('someMethod')
            ->willReturn(1);

        $this->mock
            ->expects($this->at(1))
            ->method('someMethod')
            ->willReturn(2);
    }
}
```

```php
return static function (ContainerConfigurator $containerConfigurator): void {

    $services->set(\Rector\PHPUnit\Rector\ClassMethod\MigrateAtToConsecutiveExpectationsRector::class);
};
```

# Actual Behaviour
Test is _not_ changed.

JoshuaLicense:
Closes #9.

### DIFF
--- a/src/NodeAnalyzer/ExpectationAnalyzer.php
+++ b/src/NodeAnalyzer/ExpectationAnalyzer.php
@@ -7,6 +7,7 @@ namespace Rector\PHPUnit\NodeAnalyzer;
 use PhpParser\Node\Arg;
 use PhpParser\Node\Expr;
 use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\PropertyFetch;
 use PhpParser\Node\Expr\Variable;
 use PhpParser\Node\Scalar\LNumber;
 use PhpParser\Node\Stmt\Expression;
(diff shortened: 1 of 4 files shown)


TomasVotruba:
Thank you :+1:

PR #11, opened 2021-05-08
Remove DMS from PHPUnit 8 set

JoshuaLicense:
Should DMS be included in the default PHPUnit 8 upgrade set?

Users can opt to use the DMS provided polyfill package by using the existing `phpunit80-dms.php` in addition to `phpunit80.php`.


TomasVotruba:
Makes sense :+1:

PR #12, opened 2021-05-11
Update PHPUnit 9 config set

JoshuaLicense:
Includes more of the deprecated function renames in `phpunit91.php`.

Move a single rename rule (`assertRegExp`) from `phpunit90.php` to `phpunit91.php`, but `assertRegExp` was not deprecated until PHPUnit 9.1. 

Fixed the rule's parent class: `PHPUnit\Framework\Assert` instead of `PHPUnit\Framework\TestCase`.


TomasVotruba:
Thank you :clap:

Issue #26, opened 2021-10-27
Referencing a class name in an exception message breaks refactoring to expectExceptionMessage


camporter:
`\Rector\PHPUnit\PhpDoc\PhpDocValueToNodeMapper::mapGenericTagValueNode` is used for both the expected exception and the expected exception message when refactoring php doc tags into exception methods, which opportunistically tries to use class constants.

If there happens to be `::` in the exception message, then the exception message isn't treated as a string. Would it be possible to handle exception messages (and maybe regexps) as strings always? Otherwise invalid class constants can be written and generate invalid PHP.

camporter:
@samsonasik Created #28 , thanks

samsonasik:
fixed at https://github.com/rectorphp/rector-phpunit/pull/29. I am closing it.

PR #28, opened 2021-11-09
For ExceptionAnnotationRector, add fixture that references a class in the exception message

camporter:
See #26


samsonasik:
@camporter I cherry-picked your commit at PR https://github.com/rectorphp/rector-phpunit/pull/29

PR #30, opened 2021-11-18
[PHPUnit] Using NodeTypeResolver->getType() instead of deprecated getObjectType()

samsonasik:
The `getObjectType()` is deprecated at rector-src's AbstractRector

https://github.com/rectorphp/rector-src/blob/c6473c6a0b4ecaab710353e83f0c740e2e998d26/src/Rector/AbstractRector.php#L335-L338

This PR update to use NodeTypeResolver->getType()


samsonasik:

samsonasik:
I tagged rector-phpunit 0.11.23 https://github.com/rectorphp/rector-phpunit/releases/tag/0.11.23 for it so next we can remove method `getObjectType()` from `AbstractRector` classs.

PR #36, opened 2021-12-16
[PHPUnit] Skip AssertIssetToSpecificMethodRector when parent class not detected by PHPStan

samsonasik:
Given the following code:

```php
// parent class not loaded
final class CustomIsset2 extends SomeAbstractClass {
}

final class SkipCustomIsset2 extends \PHPUnit\Framework\TestCase
{
    public function test()
    {
        $foo = new CustomIsset2();
        $this->assertTrue(isset($foo->bar));
    }
}
```

It result:

```diff
-        $this->assertTrue(isset($foo->bar));
+        $this->assertObjectHasAttribute('bar', $foo);
```

which may be invalid as `SomeAbstractClass` not found by PHPStan, which may actually exists, but not loaded early. 

This PR handle it.


samsonasik:

PR #42, opened 2022-02-01
Fix MigrateAtToConsecutiveExpectationsRector for other types

TomasVotruba:
Ref rectorphp/rector#6976

Follow up to https://github.com/rectorphp/rector-phpunit/pull/41

### DIFF
--- a/src/NodeAnalyzer/ExpectationAnalyzer.php
+++ b/src/NodeAnalyzer/ExpectationAnalyzer.php
@@ -58,14 +58,15 @@ final class ExpectationAnalyzer
                 continue;
             }
 
-            $expectsArg = $expects->args[0];
-            /** @var MethodCall $expectsValue */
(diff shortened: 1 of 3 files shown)


src/NodeAnalyzer/ExpectationAnalyzer.php
```diff
@@ -99,13 +100,17 @@ public function isValidExpectsCall(MethodCall|StaticCall $call): bool
         return count($call->getArgs()) === 1;
     }
 
-    public function isValidAtCall(MethodCall|StaticCall $call): bool
+    public function isValidAtCall(Expr $expr): bool
     {
-        if (! $this->testsNodeAnalyzer->isInPHPUnitMethodCallName($call, 'at')) {
+        if (! $expr instanceof StaticCall && ! $expr instanceof MethodCall) {
```
TomasVotruba, line 105: That was the fix :+1:

PR #44, opened 2022-02-02
[MigrateAtToConsecutiveExpectationsRector] Deprecate rule as way too ambiguous paths to upgrade and breaking, better handle manually

TomasVotruba:
After 3 hours of trying to fix the issue and looking at original PHPUnit issue with over 14 participants, https://github.com/sebastianbergmann/phpunit/issues/4297, this Rector rule took too big piece of pie.

I checked the Moodle project that upgrade their tests - https://github.com/stronk7/moodle/commit/e4f1c43ba093ed752f40749b18238f9aba222524, and there is many way to handle this upgrade. Most of them are:

```diff
// 51 changes in total
-$this->at(...);
+$this->once(...);
```

But there are couple other way, like:

```diff
// 8 changes
+$this->exactly();
```

or:

```diff
// 3 changes
+->withConsecutive(
```

Which is interesting, because the `->withConsecutive()` was actually default solution for Rector, but here it's rather edge case.

<br>

There is no clear way to upgrade this A → B. We should leave this to the end-user and PHPStorm find + replace with fixing failing test case process. It's better to handle manually to avoid random breaks.

That's why this rule is removed.

---

Closes https://github.com/rectorphp/rector/issues/6976, cc @umpirsky


umpirsky:
Too bad, but thanks for looking into it and posting detailed findings. :+1:

TomasVotruba:
Thanks for understanding 👍

PR #55, opened 2022-04-11
Fix unexpected doesNotPerformAssertions for tests with prophecy should

alexander-schranz:
fixes https://github.com/rectorphp/rector/issues/7107


samsonasik (approved):

TomasVotruba:
Thank you @alexander-schranz

PR #56, opened 2022-04-12
[DX] Use RectorConfig instead of ContainerConfigurator

samsonasik:
Ref https://github.com/rectorphp/rector-src/pull/2065#issue-1201810126

### DIFF
--- a/config/sets/phpunit91.php
+++ b/config/sets/phpunit91.php
@@ -7,7 +7,7 @@ use Rector\Renaming\Rector\MethodCall\RenameMethodRector;
 use Rector\Renaming\ValueObject\MethodCallRename;
 use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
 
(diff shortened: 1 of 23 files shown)


config/sets/phpunit91.php
```diff
@@ -2,51 +2,37 @@
...
+return static function (RectorConfig $containerConfigurator): void {
     $services = $containerConfigurator->services();
 
     $services->set(AddProphecyTraitRector::class);
 
     $services->set(RenameMethodRector::class)
         ->configure([
             // https://github.com/sebastianbergmann/phpunit/issues/4087
-            new MethodCallRename('PHPUnit\Framework\Assert', 'assertRegExp', 'assertMatchesRegularExpression'),
+            new MethodCallRename(Assert::class, 'assertRegExp', 'assertMatchesRegularExpression'),
```
samsonasik, line 20: it somehow changed to `::class` usage, which may cause issue on scoped.

PR #59, opened 2022-04-12
[DX] Use RectorConfig in tests

samsonasik:


tests/Rector/ClassMethod/AddDoesNotPerformAssertionToNonAssertingTestRector/config/configured_rule.php
```diff
@@ -2,10 +2,11 @@
 
 declare(strict_types=1);
 
+use Rector\Config\RectorConfig;
+
 use Rector\PHPUnit\Rector\ClassMethod\AddDoesNotPerformAssertionToNonAssertingTestRector;
-use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
 
-return static function (ContainerConfigurator $containerConfigurator): void {
+return static function (RectorConfig $containerConfigurator): void {
```
TomasVotruba, line 9: The variable name should reflect new type as well. 
Maybe running Rector on configs should fix it 👍

samsonasik: @TomasVotruba the `Naming` SetList can be used for it, ref https://github.com/rectorphp/rector-src/pull/2042

For note: after all packages configured with `RectorConfig`, we may need to create `Transform` Rector rule for it to change the method call.

PR #61, opened 2022-04-19
Update configs to use RectorConfig + fix assertEqualsWithDelta()

TomasVotruba:


config/sets/phpunit-code-quality.php
```diff
@@ -14,19 +14,17 @@
...
-    $services->set(ConstructClassMethodToSetUpTestCaseRector::class);
-    $services->set(AssertSameTrueFalseToAssertTrueFalseRector::class);
-    $services->set(AssertEqualsToSameRector::class);
-    $services->set(AssertCompareToSpecificMethodRector::class);
-    $services->set(AssertComparisonToSpecificMethodRector::class);
+    $rectorConfig->rule(ConstructClassMethodToSetUpTestCaseRector::class);
+    $rectorConfig->rule(AssertSameTrueFalseToAssertTrueFalseRector::class);
+    $rectorConfig->rule(AssertEqualsToSameRector::class);
+    $rectorConfig->rule(AssertCompareToSpecificMethodRector::class);
+    $rectorConfig->rule(AssertComparisonToSpecificMethodRector::class);
```
TomasVotruba, line 29: @samsonasik Could you handle the same upgrades in other packages? Should include sets and test as well.
We should use the new `rule()` methods, calls instead of parameters etc.

In the end there should be no `$parameters` nor `$rectorConfig->set(...)` :) 

Thanks

samsonasik: I will try.

PR #73, opened 2022-05-16
Rector converted assertContains(string, array) to assertStringContainsString(string, array)

tomasfejfar:
Not sure how that could happen. 

Concrete code:
```
$columns = array_map(function ($row) {
     return $row['column_name'];
}, $connection->fetchAll(sprintf('SHOW COLUMNS IN %s', $connection->quoteIdentifier('updatedDisplayName'))));

$this->assertContains('add_test_column', $columns);

// fetchAll(): array - it's a custom ODBC driver implementation. My only idea was that somehow it thought that it could be false as well 
```


tomasfejfar:
Hmmm, damn, it seems it passes as expected. I'll need to dig deeper 😞

TomasVotruba:
Have you run tests locally with different result?

TomasVotruba:
I'll merge this fixture so we have it covered in the test battery :+1:

TomasVotruba:
I've added one more fixture with `array_map()`, to be closer to your reported code:
https://github.com/rectorphp/rector-phpunit/pull/79

Issue #76, opened 2022-05-25
`GetMockRector` incorrectly renames _all_ methods named `getMock`


Blacksmoke16:
Given the following code:

```php
<?php

use Mockery;
use Mockery\Adapter\Phpunit\MockeryTestCase;

class ExampleTest extends MockeryTestCase
{
    public function test_run(): void
    {
        $someMock = $this->getMock();

        // ...
    }

    private function getMock()
    {
        return Mockery::mock('dep')
            ->expects('run')
            ->with(1, 2)
            ->getMock();
    }
}
```
The `GetMockRector` rector will alter it to be `$someMock = $this->createMock();`, which is now invalid since in this context `getMock` points to a private method, _not_ the PHPUnit mocking method.

Another context relates to Mockery exceptions. E.g.

```php
<?php

use Mockery\Adapter\Phpunit\MockeryTestCase;
use Mockery\Exception\NoMatchingExpectationException;

abstract class CustomizedTestCase extends MockeryTestCase
{
    public function runTest(): void
    {
        try {
            parent::runTest();
        } catch (NoMatchingExpectationException $ex) {
            
            /** @var Mockery\MockInterface $failedMock */
            $failedMock = $ex->getMock();

            // ...
        }
    }
}
```
Would be updated to `$failedMock = $ex->createMock();`, which of course is invalid due to that method not existing on that exception type.

TomasVotruba:
Thanks for reporting.

Fixed in https://github.com/rectorphp/rector-phpunit/pull/77

PR #77, opened 2022-05-25
Skip local getMock() method on GetMockRector

TomasVotruba:
Closes https://github.com/rectorphp/rector-phpunit/issues/76

### DIFF
--- a/src/Rector/StaticCall/GetMockRector.php
+++ b/src/Rector/StaticCall/GetMockRector.php
@@ -8,7 +8,9 @@ use PhpParser\Node;
 use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Expr\StaticCall;
 use PhpParser\Node\Identifier;
+use PHPStan\Reflection\ClassReflection;
 use Rector\Core\Rector\AbstractRector;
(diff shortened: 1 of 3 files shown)


src/Rector/StaticCall/GetMockRector.php
```diff
@@ -61,6 +83,17 @@ public function refactor(Node $node): ?Node
...
+
+        if (! $this->methodClassOwnerAnalyzer->isMethodClassOwner(
+            $callerType,
+            $scope,
+            'getMock',
+            'PHPUnit\Framework\TestCase'
+        )) {
+            return null;
+        }
+
```
samsonasik, line 96: should can be handled by `ReflectionResolver->resolveClassReflectionSourceObject()`:

```php
$classReflection = $this->reflectionResolver->resolveClassReflectionSourceObject($node);
if ($classReflection instanceof  \PHPStan\Reflection\ClassReflection && $classReflection->getName() === 'PHPUnit\Framework\TestCase') {
    return null;
}
```

TomasVotruba: Good tip, thanks :+1:

Issue #86, opened 2022-06-24
GetMockBuilderGetMockToCreateMockRector breaks test code


kenjis:
```diff
1) tests/system/Email/EmailTest.php:138

    ---------- begin diff ----------
@@ @@

     public function testDestructDoesNotThrowException()
     {
-        $email = $this->getMockBuilder(Email::class)
-            ->disableOriginalConstructor()
-            ->onlyMethods(['sendCommand'])
-            ->getMock();
+        $email = $this->createMock(Email::class);
         $email->method('sendCommand')
             ->willThrowException(new ErrorException('SMTP Error.'));
    ----------- end diff -----------

Applied rules:
 * GetMockBuilderGetMockToCreateMockRector (https://github.com/lmc-eu/steward/pull/187/files#diff-c7e8c65e59b8b4ff8b54325814d4ba55L80)
```

But when I run the test, I got the error:
```console
$ vendor/bin/phpunit tests/system/Email/EmailTest.php
PHPUnit 9.5.21 #StandWithUkraine

Runtime:       PHP 8.0.20
Configuration: .../CodeIgniter4/phpunit.xml

........E                                                           9 / 9 (100%)E                                                           9 / 9 (100%)

Time: 00:00.358, Memory: 14.00 MB

There was 1 error:

1) CodeIgniter\Email\EmailTest::testDestructDoesNotThrowException
PHPUnit\Framework\MockObject\MethodNameNotConfiguredException: Method name is not configured

ERRORS!
Tests: 9, Assertions: 20, Errors: 1.
```

samsonasik:
it seems there is test that remove `onlyMethods` for it:

https://github.com/rectorphp/rector-phpunit/blob/3f4cc815bb7cd8c5502cd873139c4cfc2734bfe5/tests/Rector/MethodCall/GetMockBuilderGetMockToCreateMockRector/Fixture/only_methods.php.inc#L11

and it registered at:

https://github.com/rectorphp/rector-phpunit/blob/3f4cc815bb7cd8c5502cd873139c4cfc2734bfe5/src/Rector/MethodCall/GetMockBuilderGetMockToCreateMockRector.php#L25

@kenjis is the error happen when `->method('sendCommand')` is called?

kenjis:
Yes, I think so.

TomasVotruba:
Let's continue discussion in one place to keep it simple: https://github.com/rectorphp/rector-phpunit/pull/87#issuecomment-1167574087

[...]

PR #87, opened 2022-06-24
[PHPUnit] Skip onlyMethods on GetMockBuilderGetMockToCreateMockRector

samsonasik:
Fixes https://github.com/rectorphp/rector-phpunit/issues/86


samsonasik:
Fixed 🎉 /cc @kenjis

samsonasik:

tests/Rector/MethodCall/GetMockBuilderGetMockToCreateMockRector/Fixture/skip_only_methods_next_call_method.php.inc
```diff
@@ -0,0 +1,17 @@
...
+
+final class SkipOnlyMethodsNextCallMethod extends \PHPUnit\Framework\TestCase
+{
+    public function testDestructDoesNotThrowException()
+     {
+        $email = $this->getMockBuilder(Email::class)
+            ->disableOriginalConstructor()
+            ->onlyMethods(['sendCommand'])
+            ->getMock();
+         $email->method('sendCommand')
```
paulbalandan, line 13: @samsonasik does this cover the test case where mock creation is in `setUp` while method invocation is in a test method? The fix assumes both mock method creation and invocation are both in the same method body whereas there may be cases where the creation is in `setUp`.

I think another fix (although not yet tested) is to check the array argument supplied to `onlyMethods` if non-empty-array. If empty, it is indeed useless.

paulbalandan: Those are already deprecated anyway so I'm not sure if there's a need.

TomasVotruba:
Could you share minimal reproducer for this? I've removed the mocks on a project event with setMethods() and tests were just fine.

samsonasik:
@TomasVotruba the example is at https://github.com/codeigniter4/CodeIgniter4/pull/6184/files#diff-c6d8f2c99fd35600ae131dea56704515c826839931d69dabb2e9cf5d38d94450 which replace with `createMock()` make error:

```bash
1) CodeIgniter\Email\EmailTest::testDestructDoesNotThrowException
PHPUnit\Framework\MockObject\MethodNameNotConfiguredException: Method name is not configured
```

@kenjis @paulbalandan I may need your help to create a minimal simple repo for it. Could you help with this? thank you.

paulbalandan:
It's weird. My minimal repo passes. But on the PR branch in CI4 on testing locally it fails. 🤔

paulbalandan:
Here's the reproducer: https://github.com/paulbalandan/create-mock
And the test run: https://github.com/paulbalandan/create-mock/runs/7054357935?check_suite_focus=true

samsonasik:
@paulbalandan thank you, I think it seems the `onlyMethods` should be skipped instead if `->onlyMethods()`, what do you think?

samsonasik:
@paulbalandan I updated to skip `onlyMethods` https://github.com/rectorphp/rector-phpunit/pull/87/commits/dd9c7635bb5b6d23d7bc1c2ed524248e63e98602

TomasVotruba:
@paulbalandan Thank you :+1: I'm not very good at mocking, so it's a challange to read.
Seems 2 cases have different configuration: https://github.com/paulbalandan/create-mock/blob/43d5161029bdfb0e6eb7cff638dab1935c46dcaf/tests/FooTest.php#L26-L38

Could you narrow th 4 methods to only 2 methods with identical code after it?

paulbalandan:
@TomasVotruba That's my mistake. Changed.

After so many simulations I think I found the cause. If the mocked method is invoked in another method, there exists a difference in using `createMock` as against `getMockBuilder()->onlyMethods()->getMock()`. In the latter case, since you specify the method(s) only to be mocked, the underlying code of the other methods are left unchanged. However, in the former case, I'm not sure if all methods were mocked. Then, when you invoke a method, if you did not a method matcher for that previously (using `->method()->with()->will()`), the invocation will fail.

TomasVotruba:
@paulbalandan Thank you :+1: Could you cut down the 4 test method to only 2? One working and other failing?

paulbalandan:
Updated.

TomasVotruba:
@paulbalandan Thank you :wink:

TomasVotruba:
@paulbalandan I looked into the test case and it seems something is wrong. Exact `createMock()` setup works for me in another project.

I cannot see why this is different.

I tried to mock returned value to avoid complexity, and the mocked method itself seems does not work:

https://github.com/paulbalandan/create-mock/pull/1

The test should pass there, right?

paulbalandan:
Based on my observations, it seems when using `createMock`, PHPUnit will mock **all** defined methods of a class. That is, each method body is changed to your configured mock logic. So, it is up to the tester to setup each method's `->method()->will(..)`.

Since in the minimal repo, we configured the mock return of only one method, the other unconfigured method has an empty body, leading to `null` return.

SantosSi:
The reason for the misbehavior is that just checking for the existence of `getMockBuilder()` without regarding the following fluent style call(s) that further configure(s) the mock's behavior it is wrong. Reason is that on PHPUnit 9.5.22 `createMock()` is an equivalent of and in the end runs:
```
       return $this->getMockBuilder($originalClassName)
                    ->disableOriginalConstructor()
                    ->disableOriginalClone()
                    ->disableArgumentCloning()
                    ->disallowMockingUnknownTypes()
                    ->getMock();
```
which is not identical to the more verbose calls it suggests to replace in the original code.
Instead the rector should only replace if the functions from the snippet above are called or, when omitted, they are the default in the PHPUnit version.
BTW, I found the URL displayed next to the applied rule just proves that someone did this, but not why and under what conditions. This PR proves that that was not enough to replace someone else's source code. I found [this](https://stackoverflow.com/questions/38363086/what-is-the-difference-between-createmock-and-getmockbuilder-in-phpunit) a lot more helpful.

samsonasik:
I am closing this for now as I have no idea to resolve it for now, the `$rectorConfig->skip()` is possibly the best way to do for this kind of `onlyMethods` on mock usecase.

SantosSi:
@samsonasik I am afraid skipping on demand would not be enough, since replacing `getMockBuilder()` calls with `createMock()` ONLY does NOT break working code under certain conditions: see the called method's equivalence in my previous post (just a matter of reading the `createMock()` and subsequently called method in PHPUnit). It is not only related to `onlyMethods()`, but already influenced on how MockBuilder's behavior is set one step before.
So the right default behavior would be to deactivate this rule unless activated, not the other way around.

samsonasik:
@SantosSi feel free to provide a patch :)

SantosSi:
@samsonasik Don't know if I can quickly read into the software structure, especially if it is for an expected smaller change like this. I found GetMockBuilderGetMockToCreateMockRector from config/sets/phpunit60.php and config/sets/remove-mocks.php and would remove it from there.
BTW, would I have the rights to do so?

samsonasik:
@SantosSi the rule purpose is to allow change `getMockBuilder()` to `createMock()`, so I prefer it still in the config set list.

if you have specific test file(s) that require the `getMockBuilder()`, I think the skip is the way to go: 

```php
$rectorConfig->skip([
        \Rector\PHPUnit\Rector\MethodCall\GetMockBuilderGetMockToCreateMockRector::class => [
            __DIR__ . '/tests/system/Email/EmailTest.php',
        ],
]);
```

or just skip entirely:

```php
$rectorConfig->skip([
        \Rector\PHPUnit\Rector\MethodCall\GetMockBuilderGetMockToCreateMockRector::class,
]);
```

SantosSi:
The main point here is that `getMockBuilder()` can ONLY be replaced by `createMock()` under specific circumstances. If these are not given, doing that the replacement breaks working code.
So, in my opinion a safe use of Rector that includes rule sets GetMockBuilderGetMockToCreateMockRector is part of either:
- has to check if the conditions are all met to safely replace the commands
- or do an opt-in instead of an opt-out.
Anyway, I have not found a reason why the replacement makes sense at all, independent of introducing the problems I explained.

JohJohan:
> The reason for the misbehavior is that just checking for the existence of `getMockBuilder()` without regarding the following fluent style call(s) that further configure(s) the mock's behavior is wrong. Reason is that on PHPUnit 9.5.22 `createMock()` is an equivalent of and in the end runs:
> 
> ```
>        return $this->getMockBuilder($originalClassName)
>                     ->disableOriginalConstructor()
>                     ->disableOriginalClone()
>                     ->disableArgumentCloning()
>                     ->disallowMockingUnknownTypes()
>                     ->getMock();
> ```
> 
^ is this not what we should check on only these functions are allowed to be called before replacement?

samsonasik:
@JohJohan I've limited knowledge of this bahaviour, could you try provide a patch? Probably collaborate with @kenjis and @paulbalandan for feedback/review. Thank you.

JohJohan:
> @JohJohan I've limited knowledge of this bahaviour, could you try provide a patch? Probably collaborate with @kenjis and @paulbalandan for feedback/review. Thank you.

Yeah i also dont know much, i create pull request and wait for feedback @SantosSi from as well

kenjis:
> createMock(string $type) returns a mock object for the specified type. The creation of this mock object is performed using best practice defaults: the __construct() and __clone() methods of the original class are not executed and the arguments passed to a method of the test double will not be cloned.
>
> If these defaults are not what you need then you can use the getMockBuilder(string $type) method to customize the test double generation using a fluent interface.
>
> https://docs.phpunit.de/en/10.2/test-doubles.html#createmock

Perhaps there are not many cases where code using `getMockBuilder()` can be refactored to `createMock()`.

JohJohan:
See https://github.com/rectorphp/rector-phpunit/pull/205 i created a merge with what i think should fix it

PR #110, opened 2022-09-13
Fix Doesnotperformassert rule by increasing the nesting level to 5

JohJohan:
Closes https://github.com/rectorphp/rector/issues/7421


TomasVotruba:
Thank you :+1:

PR #114, opened 2022-10-08
Add ProphecyPHPDocRector rule

### LINKED ISSUE #83: Rule suggestion: Using rector to fix prophecy typehints
We have a big codebase in @sulu and lots of our testcases are not really analyzeable by phpstan  as there are lot of false typehints for mocks (we have a very very big baseline). We are using `prophecy` for mocking. And there are different false typehints in our codebase:

Variant A (class but didnt define mock):

```php
class MyClassTest extends TestCase {
    /**
     * @var MyClass
     */
    private $my;

    public function setUp(): void
    {
         $this->my = $this->prophesize(MyClass::class);
    }
}
```

Variant B (mock but didnt define class):

```php
class MyClassTest extends TestCase {
    /**
     * @var ObjectProphecy
     */
    private $my;

    public function setUp(): void
    {
         $this->my = $this->prophesize(MyClass::class);
    }
}
```

Variant C (mock or class (is still false but works sometimes)):

```php
class MyClassTest extends TestCase {
    /**
     * @var ObjectProphecy|MyClass
     */
    private $my;

    public function setUp(): void
    {
         $this->my = $this->prophesize(MyClass::class);
    }
}
```

Variant D (nothing defined):

```php
class MyClassTest extends TestCase {
    private $my;

    public function setUp(): void
    {
         $this->my = $this->prophesize(MyClass::class);
    }
}
```

Expected would be ObjectProphecy Generic with the class:

```php
class MyClassTest extends TestCase {
    /**
     * @var ObjectProphecy<MyClass>
     */
    private $my;

    public function setUp(): void
    {
         $this->my = $this->prophesize(MyClass::class);
    }
}
```

So I asked myself if rector could be used to provide a rule to analyze all `$this->prophesize` calls and set the correct type for this classes. Think it would be a great rule which could be shipped with rector via a CodeQuality ruleset. Any hints are welcome how to achieve this :)

alexander-schranz:
fixes #83

### DIFF
--- a/src/Rector/Property/ProphecyPHPDocRector.php
+++ b/src/Rector/Property/ProphecyPHPDocRector.php
@@ -21,6 +21,9 @@ use Rector\PHPUnit\NodeAnalyzer\TestsNodeAnalyzer;
 use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
 use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
 
+/**
+ * @see \Rector\PHPUnit\Tests\Rector\Property\ProphecyPHPDocRector\ProphecyPHPDocRectorTest
+ */
 class ProphecyPHPDocRector extends AbstractRector
 {
     public function __construct(
@@ -129,22 +132,25 @@ CODE_SAMPLE
 
         $class = $this->betterNodeFinder->findParentType($node, Class_::class);
 
-        foreach ($class->stmts as $stmt) {
-            if (! $stmt instanceof Property) {
+        foreach ($class->stmts as $propertyStmt) {
+            if (! $propertyStmt instanceof Property) {
                 continue;
             }
 
-            if ($stmt->props[0]->name->name !== $propertyName) {
+            if ($propertyStmt->props[0]->name->name !== $propertyName) {
                 continue;
             }
 
             $doc = new Doc("/**\n     * @var ObjectProphecy<" . $prophesizeClassParts[array_key_last(
                 $prophesizeClassParts
             )] . ">\n     */");
-            $stmt->setDocComment($doc);
-            $stmt->getDocComment();
+            $propertyStmt->setDocComment($doc);
+            $propertyStmt->getDocComment();
 
-            $useStatements = ['Prophecy\Prophecy\ObjectProphecy', \implode('\\', $prophesizeClassParts)];
+            $useClassNames = [
+                'Prophecy\Prophecy\ObjectProphecy',
+                \implode('\\', $prophesizeClassParts),
+            ];
 
             $namespace = $this->betterNodeFinder->findParentType($node, Namespace_::class);
 
@@ -152,20 +158,20 @@ CODE_SAMPLE
                 return;
             }
 
-            foreach ($namespace->stmts as $stmt) {
-                if (! $stmt instanceof Use_) {
+            foreach ($namespace->stmts as $useStmt) {
+                if (! $useStmt instanceof Use_) {
                     continue;
                 }
 
-                $foundClass = \implode('\\', $stmt->uses[0]->name->parts);
+                $foundClass = \implode('\\', $useStmt->uses[0]->name->parts);
 
-                $findIndex = \array_search($foundClass, $useStatements, true);
+                $findIndex = \array_search($foundClass, $useClassNames, true);
                 if ($findIndex !== false) {
-                    unset($useStatements[$findIndex]);
+                    unset($useClassNames[$findIndex]);
                 }
             }
 
-            $uses = $this->nodeFactory->createUsesFromNames($useStatements);
+            $uses = $this->nodeFactory->createUsesFromNames($useClassNames);
             $stmts = $namespace->stmts;
 
             foreach ($uses as $use) {
--- a/src/Rector/Property/ProphecyPHPDocRector.php
+++ b/src/Rector/Property/ProphecyPHPDocRector.php
@@ -1,0 +1,180 @@
+<?php
+
+namespace Rector\PHPUnit\Rector\Property;
+
+use PhpParser\Comment\Doc;
+use PhpParser\Node\Expr\Assign;
+use PhpParser\Node\Expr\ClassConstFetch;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\PropertyFetch;
+use PhpParser\Node\Expr\Variable;
+use PhpParser\Node\Identifier;
+use PhpParser\Node\Scalar\String_;
+use PhpParser\Node\Stmt\Class_;
+use PhpParser\Node\Stmt\Namespace_;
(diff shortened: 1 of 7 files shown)


src/Rector/Property/ProphecyPHPDocRector.php
```diff
@@ -0,0 +1,184 @@
...
+            }
+
+            $uses = $this->nodeFactory->createUsesFromNames($useClassNames);
+            $stmts = $namespace->stmts;
+
+            foreach ($uses as $use) {
+                $stmts[] = $use; // TODO not working currently
+            }
+
+            $namespace->stmts = $stmts;
```
alexander-schranz, line 179: this lines fails which make the current tests failing.

src/Rector/Property/ProphecyPHPDocRector.php
```diff
@@ -0,0 +1,184 @@
...
+        if ($name->name !== 'prophesize') {
+            return;
+        }
+
+        $value = $expr->args[0]->value;
+
+        if ($value instanceof String_) {
+            $prophesizeClassParts = \explode('\\', $value->value);
+        } elseif ($value instanceof ClassConstFetch) {
+            $prophesizeClassParts = $value->class->parts;
```
alexander-schranz, line 121: While appling the rule on my codebase: This current fails in case of an alias import of the class e.g.:

```php
use Some\A as B;
use Some\Other\A;

$this->test = $this->prophesize(B::class);
```

Ends in:

```
/** @var ObjectProphecy<A> */
private $test;
```

Should be:

```
/** @var ObjectProphecy<B> */
private $test;
```

test required for this case.

src/Rector/Property/ProphecyPHPDocRector.php
TomasVotruba: By convention, we keep all class names FQN. Coding standard tool handles this, or `$rectorConfig->importNames();`.
In short, you never have to deal with FQN in any rule :)

src/Rector/Property/ProphecyPHPDocRector.php
TomasVotruba: Same here, always use FQN class name.

src/Rector/Property/ProphecyPHPDocRector.php
alexander-schranz: PHP-CS-Fixer and importNames did not work in this case, think they can not handle this kind of phpdocs.

I'm still curious what would be the correct way to add `use` statements via rector?

TomasVotruba: In that case it should be fixed in the importNames(). This rule should not handle this.

TomasVotruba: > I'm still curious what would be the correct way to add use statements via rector?

Check the `NameImportingPostRector`  :) that's the rule that handles it.

src/Rector/Property/ProphecyPHPDocRector.php
```diff
@@ -0,0 +1,184 @@
...
+}
+CODE_SAMPLE
+                    ,
+                    <<<'CODE_SAMPLE'
+class HelloTest extends TestCase
+{
+    /**
+     * @var ObjectProphecy<SomeClass>
+     */
+    private $test;
```
TomasVotruba, line 62: We can take it further. If on PHP 7.4+, this should be typed property, right? :)

src/Rector/Property/ProphecyPHPDocRector.php
```diff
@@ -0,0 +1,184 @@
...
+                ),
+            ]
+        );
+    }
+
+    public function getNodeTypes(): array
+    {
+        // PhpParser\Node\Expr\Assign
+        // PhpParser\Node\Expr\MethodCall
+        return [Assign::class];
```
TomasVotruba, line 79: The best practise while working with AST is to hook into the node that is being changed. 
Otherwise you might try to change nodes, that were already traversed and changed.

Here you change the `Property` doc/type, so you should hook into `Class_` node :+1:

tests/Rector/Property/ProphecyPHPDocRector/config/configured_rule.php
```diff
@@ -0,0 +1,15 @@
...
+use Rector\Config\RectorConfig;
+use Rector\PHPUnit\Rector\Property\ProphecyPHPDocRector;
+
+return static function (RectorConfig $rectorConfig): void {
+    $rectorConfig->import(__DIR__ . '/../../../../../config/config.php');
+
+    $rectorConfig->rule(ProphecyPHPDocRector::class);
+
+    $rectorConfig->importNames();
+    $rectorConfig->importShortClasses(false);
```
TomasVotruba, line 14: By convention, the test should only verify the rule alone, without any configuration.

The imports can be handled in a follow up PR or better in the import rule itself.

src/Rector/Property/ProphecyPHPDocRector.php
alexander-schranz: Converting phpdoc to type property is an own rule. Which can just be applied if somebody want it.

src/Rector/Property/ProphecyPHPDocRector.php
alexander-schranz: I will not be able todo this in the next time, if somebody want to take it over and convert it to that I'm fine.

tests/Rector/Property/ProphecyPHPDocRector/config/configured_rule.php
alexander-schranz: I just added them to show that they are not working like expected ;).

I repushed and update the tests. If somebody want to take it over to convert it to `Class_` or `Property` listener I'm fine with it, I will not able todo that soon.

I'm still curious about the `use statement` part what is the correct way to add any use statements to a file?

TomasVotruba: See https://github.com/rectorphp/rector-phpunit/pull/114#discussion_r990774677

TomasVotruba:
Thanks 👍 seems almost finished 🎉

Do you need help with PHPStan or can you handle it?

src/Rector/Property/ProphecyPHPDocRector.php
alexander-schranz: Okay looks like a postprocessor is adding that I understand. Thx for the insights.

TomasVotruba: 👍

TomasVotruba:
I've cherry picked your rule and fixed the PHPStan errors here:
https://github.com/rectorphp/rector-phpunit/pull/115

TomasVotruba:
Thank you for the work :+1: :clap:

alexander-schranz:
@TomasVotruba Thx for fixing the open issues 👍

TomasVotruba:
👍

PR #115, opened 2022-10-09
Add ProphecyPHPDocRector rule - on the Class_ node

TomasVotruba:


TomasVotruba:
@todo do not use strings in Doc, but objects

alexander-schranz:
@TomasVotruba was this rule renamed? Currently searching for it.

TomasVotruba:
See https://github.com/search?q=repo%3Arectorphp%2Frector-phpunit+ProphecyPHPDocRector&type=pullrequests

It was part of docblock-rules removal, as since Rector 0.15 we're moving away form any docblock changes/upgrades as not reliable: https://getrector.com/blog/new-in-rector-015-complete-safe-and-known-type-declarations

I'd recommend to move away from mocks completelly with anonymous class to get full power of PHPStan and Rector.

alexander-schranz:
Thx for the hint. The code base is too big to change away from mocking. Also when I completly fan of no mocks and even would prefer to have InMemoryRepositories for all my entities, but world not always perfect.

Also no PHPDoc does not work in case of this Generics as PHP does not yet supports for Generics . And to work correctly `ObjectProphecy<MyInterface>` is required to get PHPStan work correctly on this tests :).

Maybe I will move then the rule into an own repository if it still possible to achieve this with rector.

TomasVotruba:
> The code base is too big to change away from mocking.

The size doesn't matter with Rector :) we're successfuly moving projects from mocks, unless some 3rd party code like Aws. Typically only public methods are mocked, that Rector can find and convert to anonymous class or standalone class. This also make any future changes and maintenance cheaper and easier.

PR #116, opened 2022-10-09
Use PhpDocInfo for docblock changes in ProphecyPHPDocRector

TomasVotruba:


src/Rector/Class_/ProphecyPHPDocRector.php
```diff
@@ -160,14 +162,17 @@ private function findAssignToPropertyName(Class_ $class, string $propertyName):
...
+
+        $propertyPhpDocInfo->addTagValueNode(new VarTagValueNode(
+            new GenericTypeNode(new FullyQualifiedIdentifierTypeNode('Prophecy\Prophecy\ObjectProphecy'), [
+                new FullyQualifiedIdentifierTypeNode($prophesizedClass),
+            ]),
+            '',
+            ''
+        ));
     }
 }
```
TomasVotruba, line 178: @alexander-schranz Just a tip for work with docblocks :) 
It's better to use object API with PhpDocInfo rather than strings. Rector and PHPStan can manage it better and work with those as types :+1:

alexander-schranz: Thx for the hint 👍

Issue #69, opened 2022-05-05
[PHPUnit 10] Add rule for migrate annotation to attribute


lyrixx:
ref: https://github.com/sebastianbergmann/phpunit/issues/4502

TomasVotruba:
That looks like a lot of work. We're now focusing on improving Rector core to make ruls behavior more transparent.
Would you like to contribute this one?

lyrixx:
I don't have the bandwidth 😬 I'm sorry. I opened the issue to give some visibility to this new feature of PHPUnit

TomasVotruba:
I see. We use issue to handle mostly bugs or PHP upgrade paths. Tools mostly depends on community activity, so it reflects the real demand for upgrade. 

E.g. I have used maybe 1-2 annotations from thta whole list. Let's wait for the first contributor :)

TomasVotruba:
Mostly handled now, let us know if you find a missing case!

PR #123, opened 2022-11-13
[WIP][Set] Annotation to attribute

wkania:
PHPUnit will support [attributes](https://github.com/sebastianbergmann/phpunit/issues/4502) in version 10.  
It would be cool to have day 1 support.
ref issue: https://github.com/rectorphp/rector-phpunit/issues/69

This is work in progress because currently `AnnotationToAttribute`:
1. do not support syntax like: `@dataProvider  data`  or `@group api` (those are probably most common)
2. are not aware if the annotation is attached to method or class

Probably some new class like `PHPUnitAnnotationToAttribute` is needed.
@TomasVotruba Any feedback is welcome.

### DIFF
--- a/config/sets/annotations-to-attributes.php
+++ b/config/sets/annotations-to-attributes.php
@@ -1,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+
+use Rector\Config\RectorConfig;
+use Rector\Php80\Rector\Class_\AnnotationToAttributeRector;
+use Rector\Php80\ValueObject\AnnotationToAttribute;
+
+return static function (RectorConfig $rectorConfig): void {
+    $rectorConfig->ruleWithConfiguration(AnnotationToAttributeRector::class, [
+        // @see https://github.com/sebastianbergmann/phpunit/issues/4502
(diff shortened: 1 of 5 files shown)


TomasVotruba:
Hi @wkania ,

thank you for this PR :+1: 

I've fixed issues with our CI, that was false positive fail. 


> 1. do not support syntax like: @dataProvider data or @group api (those are probably most common)

This will require a new Rector rule here to handle this, as annotation to attribute is mainly class-based now and only bare tags like `@after` are supported.


> 2. are not aware if the annotation is attached to method or class

At the moment the rule convert annotation on class on attribute on class, etc. If you need to move e.g. annotation on class to attributes on method, that will require new Rector rule too.

<br>

I think this is a good place to kick off, as we have the basic cases covered. I'll merge this so we can iterate on the rest of cases with better narrow focus :muscle:

TomasVotruba:
Here is the follow up with annotation with value:

* https://github.com/rectorphp/rector-phpunit/pull/129
* https://github.com/rectorphp/rector-phpunit/pull/130

TomasVotruba:
* https://github.com/rectorphp/rector-phpunit/pull/131

PR #124, opened 2022-11-14
Trigger CI to test Rector external workflow only for repository owners

TomasVotruba:


.github/workflows/rector.yaml
```diff
@@ -0,0 +1,15 @@
+name: Code Analysis
```
samsonasik, line 1:
```suggestion
name: Rector
```

TomasVotruba: Thanks.

I'm trying to figure out, why the workflow is not triggered bellow :/ any ideas?

samsonasik:
@TomasVotruba please add an unused private property in a class to test that rector can auto-run rectify and trigger Ci again.

eg: https://github.com/rectorphp/rector-symfony/pull/121/commits/0206ad61d0af60b01f0101e5f988fb42a9f5bf38

In the past, it run rectify, but stop the CI, see  https://github.com/rectorphp/rector-symfony/pull/121

TomasVotruba:
@samsonasik Added :+1: I think the workflow itself is not even triggered, so the CI job will not even run

.github/workflows/rector.yaml
samsonasik: I tried in the past, it apply rectify, but after rectify, it stop, see:

https://github.com/rectorphp/rector-symfony/pull/121

I added unused private property:

https://github.com/rectorphp/rector-symfony/pull/121/commits/0206ad61d0af60b01f0101e5f988fb42a9f5bf38

It then run rectify:

https://github.com/rectorphp/rector-symfony/pull/121/commits/951fec105425bf17f4452e5402db25cf3f74818a

then stop as pending:

Screen Shot 2022-03-08 at 17 07 36

TomasVotruba:
It seems removing the reuqirement helped to trigger it: https://github.com/rectorphp/reusable-workflows/commit/2f8c012ea7e239fa1ea2487c04aea326f41af383

samsonasik:
@TomasVotruba what I meant is, let `Rectify` run by CI, then see if after rectify, it pending or continue working, in the past, this is what happen:

- Rectify run: committed by github action
- CI stop

Screen Shot 2022-03-08 at 17 07 36

PR #125, opened 2022-11-14
try trigger rector

TomasVotruba:


TomasVotruba:
@samsonasik It's not perfect yet, but Rector at least triggers and commits :+1:

samsonasik:
No, in the past, it trigger and commit, but CI then stop, that require close and re-open to continue

PR #129, opened 2022-12-08
[PHPUnit 10] Add AnnotationWithValueToAttributeRector

TomasVotruba:
- cleanup
- fix level 10 set
- Add AnnotationWithValueToAttributeRector


TomasVotruba:
@sebastianbergmann Hi Sebastian, 
I've extended PHPUnit annotation to attribute set with the few edge cases, so we're ready when PHPUnit 10 is out :+1:

PR #130, opened 2022-12-08
[PHPUnit 10] Extends annotation to attribute set

TomasVotruba:
* Ref https://github.com/rectorphp/rector-phpunit/issues/69
* Ref https://github.com/sebastianbergmann/phpunit/issues/4502

Follow up to https://github.com/rectorphp/rector-phpunit/pull/129


TomasVotruba:
@sebastianbergmann Hi Sebastian, 
I've extended PHPUnit annotation to attribute set with the few edge cases, so we're ready when PHPUnit 10 is out :+1:

PR #136, opened 2022-12-17
[PHPUnit] Fix @depends on class should use DependsOnClass attribute

samsonasik:
@depends on class should use `PHPUnit\Framework\Attributes\DependsOnClass`


samsonasik:

PR #140, opened 2023-01-05
Remove unused classes

JohJohan:
It was used in TestClassResolverTest but that is not in the project anymore and I don't see these classes used


TomasVotruba:
Thank you!

Could you fix the Rector build as well?

samsonasik:
Thank you @JohJohan

Issue #142, opened 2023-01-11
Data providers may come from extended classes


marmichalski:
Imagine a Test class, let's call it A, extending some other (potentially) abstract Test class, B, that has defined example dataProvider methods and class A is using data providers from class B.  
This makes [DataProviderClassMethodFinder](https://github.com/rectorphp/rector-phpunit/blob/97cab8de4f7c6cd2b1006e3394409a7c98425bde/src/NodeFinder/DataProviderClassMethodFinder.php#L31) throw exception as the docblock referenced method is not contained in the same class.

The simplest solution would be to change exception throwing into a `continue` step, disadvantage is that the extended class method won't be recognized as a data provider - but this is already happening for the base class that provides the data provider methods only 🙈.

<details>
  <summary>Very basic code sample</summary>

```php

abstract class SomeAbstractTest extends TestCase
{
    public function provideData()
    {
        yield [1];
    }
}

final class SomeTest extends SomeAbstractTest
{
    /**
     * @dataProvider provideData()
     */
    public function test()
    {
    }
}
```
</details>

TomasVotruba:
Resolved in https://github.com/rectorphp/rector-phpunit/pull/143

Thank you :+1:

PR #143, opened 2023-01-11
Ignore non existent data provider methods

marmichalski:
This PR, when merged, would unblock the ~failures~ failure described in #142.

Handling non-final classes that define data provider methods and not use them themself could be handled as a feature, I guess.


samsonasik (approved):

TomasVotruba:
Thanks :+1:

PR #144, opened 2023-01-12
Update assertion name for AssertTrueFalseToSpecificMethodRector

GeniJaho:
Fixes https://github.com/rectorphp/rector/issues/7626.


samsonasik:
is this tied to specific start from version phpunit? eg: it start from phpunit 9.1, then it needs to be in phpunit91 config set.

GeniJaho:
There is actually a rule in the phpunit91 set that renames the method to assertFileDoesNotExist, so this change is not even needed, probably?

samsonasik:
Probably just keep as is, and add new `MethodCallRename` in phpunit91 config:

```php
        new MethodCallRename('PHPUnit\Framework\Assert', 'assertFileNotExists', 'assertFileDoesNotExist'),
```

GeniJaho:
It's already there 😁 so I'm closing this PR.
https://github.com/rectorphp/rector-phpunit/blob/main/config/sets/phpunit91.php#L19

PR #147, opened 2023-02-03
Update to PHPUnit 10

TomasVotruba:
- easy ci
- bump


src/Rector/ClassMethod/DependsAnnotationWithValueToAttributeRector.php
```diff
@@ -159,11 +162,11 @@ private function resolveAttributeValueAndAttributeName(
             $originalAttributeValue
         );
 
-        $attributeName = 'PHPUnit\Framework\Attributes\Depends';
+        $attributeName = Depends::class;
```
samsonasik, line 165: this will be prefixed, so it should not use `::class`

src/Rector/Class_/AnnotationWithValueToAttributeRector.php
```diff
@@ -64,7 +65,7 @@ final class SomeTest extends TestCase
 CODE_SAMPLE
                 ,
                 [
-                    new AnnotationWithValueToAttribute('backupGlobals', 'PHPUnit\Framework\Attributes\BackupGlobals', [
+                    new AnnotationWithValueToAttribute('backupGlobals', BackupGlobals::class, [
```
samsonasik, line 68: this will be prefixed, so it should not use ::class

src/Rector/Class_/CoversAnnotationWithValueToAttributeRector.php
```diff
@@ -111,10 +113,10 @@ public function refactor(Node $node): ?Node
             $annotationValue = $desiredTagValueNode->value->value;
 
             if (str_starts_with($annotationValue, '::')) {
-                $attributeClass = 'PHPUnit\Framework\Attributes\CoversFunction';
+                $attributeClass = CoversFunction::class;
```
samsonasik, line 116: this will be prefixed, so it should not use ::class

PR #154, opened 2023-02-07
Add failing test fixture for `StaticDataProviderClassMethodRector`

OskarStark:
# Failing Test for StaticDataProviderClassMethodRector

Based on https://getrector.com/demo/3074e792-c0f4-4db7-bdd4-fb2c4b55bee3

### DIFF
--- a/tests/Rector/Class_/StaticDataProviderClassMethodRector/Fixture/test_case.php.inc
+++ b/tests/Rector/Class_/StaticDataProviderClassMethodRector/Fixture/test_case.php.inc
@@ -1,0 +1,77 @@
+<?php
+
+namespace PHPUnit\Framework;
+
+namespace Rector\Tests\PHPUnit\Rector\Class_\StaticDataProviderClassMethodRector\Fixture;
+
+class TestCase
+{
+}
+
+namespace Rector\Tests\PHPUnit\Rector\Class_\StaticDataProviderClassMethodRector\Fixture;
+
+abstract class AbstractFooTestCase extends TestCase
+{
+    /**
+     * @dataProvider dataProvider
+     */
+    public function testWithDataProvider($value1, $value2): void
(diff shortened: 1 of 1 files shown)


TomasVotruba:
This should be doable, as we know about the parent class.
Please, don't fix anything manually yet... :D I'll check it today.

tests/Rector/Class_/StaticDataProviderClassMethodRector/Fixture/test_case.php.inc
```diff
@@ -0,0 +1,77 @@
+<?php
+
+namespace PHPUnit\Framework;
+
+namespace Rector\Tests\PHPUnit\Rector\Class_\StaticDataProviderClassMethodRector\Fixture;
+
+class TestCase
```
OskarStark, line 7: Not sure this works, I think you check for this namespace in rules, right?

TomasVotruba: Yeah, it should be spit 1 class = 1 file, so PHPStan can autoload it propperly.
But don't worry about it :)

OskarStark: Would be cool if one could "Add new file" like for gists.github.com on the demo page

TomasVotruba: I think that might lead people to commiting /vendor here :D 

Also, does Github link allow multiple files?

OskarStark: > Also, does Github link allow multiple files?

I don't know

PR #158, opened 2023-02-07
Add PropertyExistsWithoutAssertRector

jrjohnson:
Removes the [now deprecated](https://github.com/sebastianbergmann/phpunit/issues/4601) property/method exists PHPUnit assertions and replaces them with plain PHP equivalents. This is the inverse of [AssertPropertyExistsRector.php](https://github.com/rectorphp/rector-phpunit/blob/ee461b14551189faca960058ef4fb1146755f8ca/src/Rector/MethodCall/AssertPropertyExistsRector.php) to account for their removal in PHPUnit v10.

### DIFF
--- a/docs/rector_rules_overview.md
+++ b/docs/rector_rules_overview.md
@@ -715,6 +715,28 @@ Turns getMock*() methods to `createMock()`
 
 <br>
 
+<br>
+
+## PropertyExistsWithoutAssertRector
+
+Turns PHPUnit TestCase `assertObjectHasAttribute` into `property_exists` comparisons
+
+- class: [`Rector\PHPUnit\Rector\MethodCall\PropertyExistsWithoutAssertRector`](../src/Rector/MethodCall/PropertyExistsWithoutAssertRector.php)
+
+```diff
+-$this->assertClassHasAttribute("property", "Class");
+-$this->assertClassNotHasAttribute("property", "Class");
+-$this->assertObjectNotHasAttribute("property", $object);
+-$this->assertObjectHasAttribute("property", $object);
+
++$this->assertFalse(property_exists(new Class, "property"));
++$this->assertTrue(property_exists(new Class, "property"));
++$this->assertFalse(property_exists($object, "property"));
++$this->assertTrue(property_exists($object, "property"));
+```
+
+<br>
+
 ## ProphecyPHPDocRector
 
 Add correct `@var` to ObjectProphecy instances based on `$this->prophesize()` call.
--- a/src/Rector/MethodCall/PropertyExistsWithoutAssertRector.php
+++ b/src/Rector/MethodCall/PropertyExistsWithoutAssertRector.php
@@ -1,0 +1,126 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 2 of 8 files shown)


TomasVotruba:
I'll look into this, thank you :+1:

eerison:
Hey @TomasVotruba 

I guess this rule should be introduced in phpunit 9 rules instead of 10, don't you?

https://github.com/rectorphp/rector-phpunit/pull/202/files#diff-1d1786275faa487b46d3179a731083128edc0bb6238271e77c70c5c0ccf1881e

TomasVotruba:
Hey, the PHPUnit 10 is fine here as that's where the change happens. See milestone in: https://github.com/sebastianbergmann/phpunit/issues/4601

The idea is that after running PHPUNIT_100 set, the code should run on PHPUnit 10.

eerison:
But I can not run `PHPUNIT_100` in phpunit 9, and it is deprecated in version 9, IMO it could be in version 9, my pipelines broken when I add deprecated methods :(

the same thing for https://github.com/rectorphp/rector-phpunit/blob/main/docs/rector_rules_overview.md#addprophecytraitrector

eerison:
well I could make this work adding

```php
    $rectorConfig->rule(PropertyExistsWithoutAssertRector::class);
    $rectorConfig->rule(AddProphecyTraitRector::class);
```

But if it's in version 9, I could avoid search about those 2 rules.

TomasVotruba:
You can run `PHPUNIT_100` set on any PHPUnit version, then upgrade to PHPUnit 10 in your composer. 

What exact error are you getting?

eerison:
I saw that it is applying php 8 code, and I'm still on php 7.4 :'(

```
                return match ($matcher->numberOfInvocations()) {
                    1 => $expectedSleepValues,
                };
```

eerison:
before

```php
->withConsecutive(...$expectedSleepValues);
```

after

```php
            ->willReturnCallback(function () use ($matcher) {
                return match ($matcher->numberOfInvocations()) {
                    1 => $expectedSleepValues,
                };
            });
```

eerison:
Ok it's the issue

https://github.com/rectorphp/rector-phpunit/blob/main/src/Rector/StmtsAwareInterface/WithConsecutiveRector.php

this rule is applying php 8 code

TomasVotruba:
That's the pain point. Perfect! This rule should run only on PHP 8 indeed.

> Should we downgrade this code to php 7 ?

That's not necessary. The `withConsecutive()` still runs on PHPUnit 9, and PHPUnit 10 already requires on PHP 8.1, so it's supported there.

Instead, this rule should implement `MinPhpVersionInterface`

eerison:
> That's the pain point. Perfect! This rule should run only on PHP 8 indeed.
> 
> > Should we downgrade this code to php 7 ?
> 
> That's not necessary. The `withConsecutive()` still runs on PHPUnit 9, and PHPUnit 10 already requires on PHP 8.1, so it's supported there.
> 
> Instead, this rule should implement `MinPhpVersionInterface` with PHP 8. Could you add it?

Yep, I will open a PR ;)

PR #159, opened 2023-02-09
remove abstract *Test source file after renaming to *TestCase

arderyp:
https://github.com/rectorphp/rector/issues/7767

https://github.com/rectorphp/rector/issues/7767#issuecomment-1424577054

### DIFF
--- a/src/Rector/ClassLike/RemoveTestSuffixFromAbstractTestClassesRector.php
+++ b/src/Rector/ClassLike/RemoveTestSuffixFromAbstractTestClassesRector.php
@@ -8,6 +8,7 @@ use PhpParser\Node;
 use PhpParser\Node\Identifier;
 use PhpParser\Node\Stmt\Class_;
 use PhpParser\Node\Stmt\Namespace_;
+use Rector\Core\Application\FileSystem\RemovedAndAddedFilesCollector;
 use Rector\Core\Configuration\RenamedClassesDataCollector;
 use Rector\Core\Exception\ShouldNotHappenException;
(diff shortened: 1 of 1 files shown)


arderyp:
I am not sure if new test coverage is needed.  I see there is a test to ensure that the new/renamed file was added, but I am not sure how these fixtures work and if an actual deletion is executed and, if so, how to test it.

arderyp:
@samsonasik

samsonasik (approved):
Looks good

samsonasik:
I verify locally and it seems working ok, @TomasVotruba I am merging it ;)

Thank you @arderyp

arderyp:
very cool!

@samsonasik, in testing the rector set, I've noticed a deprecation/change that is not handled.  PHPUnit 10 seems to have done away with the old hooks (example `BeforeFirstTestHook`) and says we should now use the new `Event` system.  [This is not properly documented yet](https://github.com/sebastianbergmann/phpunit/issues/4676#issuecomment-1424814440), so I am not exactly sure what Rector should even do in these scenarios, but thought I should mention it to you.

Should I open a new issue for that? I may be able to work on it myself, but I am completely new to Rector, so it may be very slow progress.

TomasVotruba:
Thanks for looking into this. 
Go for the pull-request :+1: We'll help you to improve the rule in small iterations.

arderyp:
commenting here for when I look back later.

Here is a great write-up/overview of the new event system, which will come in handy here:

https://localheinz.com/articles/2023/02/14/extending-phpunit-with-its-new-event-system/

PR #160, opened 2023-02-11
Add failing test fixture for `StaticDataProviderClassMethodRector` with ignored abstract parent test case

OskarStark:
# Failing Test for StaticDataProviderClassMethodRector

Based on https://getrector.com/demo/9372e294-4277-4bb9-abeb-34764c0b053c

# Explanation

Because the parent abstract class is skipped in rector config, their child classes should be ignored for this rector as well. Check the demo link please, thanks.

This test should be failing ❗ 

Spotted in
* https://github.com/symfony/symfony/pull/48668

### DIFF
--- a/tests/Rector/Class_/StaticDataProviderClassMethodRector/Fixture/test_case.php.inc
+++ b/tests/Rector/Class_/StaticDataProviderClassMethodRector/Fixture/test_case.php.inc
@@ -1,0 +1,38 @@
+<?php
+
+namespace PHPUnit\Framework;
+
+namespace Rector\Tests\PHPUnit\Rector\Class_\StaticDataProviderClassMethodRector\Fixture;
+
+class TestCase
+{
+}
+
+namespace Rector\Tests\PHPUnit\Rector\Class_\StaticDataProviderClassMethodRector\Fixture;
+
+abstract class AbstractFooTestCase extends TestCase
+{
+    /**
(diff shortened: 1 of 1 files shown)


TomasVotruba:
This is identical to already fixed https://github.com/rectorphp/rector-phpunit/pull/156

How is it different?

OskarStark:
CleanShot 2023-02-11 at 13 23 22@2x

I told rector to skip the abstract class, which cannot be reflected in the failing test case

TomasVotruba:
I see. It only works for explicit paths, no magic :)

Also, the `__DIR__ . 'String` might result into `SomePathString`

TomasVotruba:
Btw, what is reason to ignore any files here? The upgrade must handle all files.

OskarStark:
In Symfony all abstract cases in /Test/ namespace are under BC promise and can be used in username. All abstract testcases in /Tests/ namespace are considered internal and are only used for our codebase, which we can change but the ones in /Test/ need to have an upgrade path including deprecations we need to find manually.

But ofc our own tests in /Test/ are using the abstract testcases from /Test/

I hope I make it clear 😃 thanks for your time

TomasVotruba:
I see :)  That's super specific use case that would be only useful for Symfony core project.

I think for that would be best just to fork Rector rule to custom one and apply this logic there.

PR #163, opened 2023-02-27
Fix for case sensitive PHPUnit annotations

tarlepp:
Fixes #7804


TomasVotruba:
Neat :) thank you :+1:

PR #167, opened 2023-04-15
Use `#[DataProviderExternal]` instead of `#[DataProvider]` attribute

staabm:
closes https://github.com/rectorphp/rector/issues/7876

### DIFF
--- a/src/Rector/ClassMethod/DataProviderAnnotationToAttributeRector.php
+++ b/src/Rector/ClassMethod/DataProviderAnnotationToAttributeRector.php
@@ -119,7 +119,10 @@ CODE_SAMPLE
             }
 
             if ($className !== '') {
-                if ($className)
+                if ($className[0] !== '\\') {
+                    $className = '\\' . $className;
+                }
+
                 $attributeGroup = $this->phpAttributeGroupFactory->createFromClassWithItems(
                     'PHPUnit\Framework\Attributes\DataProviderExternal',
                     [$className.'::class', $methodName]
--- a/src/Rector/ClassMethod/DataProviderAnnotationToAttributeRector.php
+++ b/src/Rector/ClassMethod/DataProviderAnnotationToAttributeRector.php
@@ -113,9 +113,13 @@ CODE_SAMPLE
 
             $methodName = trim($originalAttributeValue, '()');
 
+            $className = '';
             if (str_contains($methodName, '::')) {
                 [$className, $methodName] = explode('::', $methodName, 2);
+            }
 
+            if ($className !== '') {
+                if ($className)
                 $attributeGroup = $this->phpAttributeGroupFactory->createFromClassWithItems(
                     'PHPUnit\Framework\Attributes\DataProviderExternal',
                     [$className.'::class', $methodName]
--- a/src/Rector/ClassMethod/DataProviderAnnotationToAttributeRector.php
+++ b/src/Rector/ClassMethod/DataProviderAnnotationToAttributeRector.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Rector\PHPUnit\Rector\ClassMethod;
 
 use PhpParser\Node;
+use PhpParser\Node\AttributeGroup;
 use PhpParser\Node\Stmt\Class_;
 use PhpParser\Node\Stmt\ClassMethod;
 use PHPStan\PhpDocParser\Ast\PhpDoc\GenericTagValueNode;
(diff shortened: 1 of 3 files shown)


tests/Rector/ClassMethod/DataProviderAnnotationToAttributeRector/Fixture/external-provider.php.inc
```diff
@@ -0,0 +1,47 @@
...
+-----
+<?php
+
+namespace Rector\PHPUnit\Tests\Rector\ClassMethod\DataProviderAnnotationToAttributeRector\Fixture;
+
+use PHPUnit\Framework\TestCase;
+
+final class FooTest extends TestCase
+{
+    #[\PHPUnit\Framework\Attributes\DataProviderExternal(Tests\Foo\ExternalProvider::class, 'barProvider')]
```
staabm, line 34: @samsonasik atm the rector creates
```
#[\PHPUnit\Framework\Attributes\DataProviderExternal('Tests\Foo\ExternalProvider::class', 'barProvider')]
```

but we need a plain class-string without single quotes arround

```
#[\PHPUnit\Framework\Attributes\DataProviderExternal(Tests\Foo\ExternalProvider::class, 'barProvider')]
```

do you have an idea on how to create the annotation arg properly?

samsonasik: Using ::class will need FQCN classname with `\` prefix, otherwise, that refer to current namespace, except, short name that already in use statement

staabm: thx for the hint. my initial question was more about how I could make the rector produce a class-string (which is not sorrounded by single quotes)

samsonasik: quoted string should can use String_ node on attribute arg

staabm: > Using ::class will need FQCN classname with `\` prefix, otherwise, that refer to current namespace, except, short name that already in use statement

I think we need to assume that pre-rector-run existing code either correctly had a use statement or referenced using a FQCN. the rector itself does not change that class

samsonasik: if I remember correctly, using annotation `@dataProvider Foo\Class` is same with `@@dataProvider \Foo\Class`, while attribute, it processsed by PHP so : 

```php
DataProviderExternal(Tests\Foo\ExternalProvider::class, 'barProvider')]
```

vs 


```php
DataProviderExternal(\Tests\Foo\ExternalProvider::class, 'barProvider')]
```

are different.

samsonasik:
@OskarStark could you verify this? Thank you.

tests/Rector/ClassMethod/DataProviderAnnotationToAttributeRector/Fixture/external-provider.php.inc
staabm: ok - added a separate test case for your example

src/Rector/ClassMethod/DataProviderAnnotationToAttributeRector.php
```diff
@@ -129,4 +124,30 @@ public function refactor(Node $node): ?Node
 
         return $node;
     }
+
+    private function createAttribtueGroup(string $originalAttributeValue): AttributeGroup
```
samsonasik, line 128: typo:

```suggestion
    private function createAttributeGroup(string $originalAttributeValue): AttributeGroup
```

src/Rector/ClassMethod/DataProviderAnnotationToAttributeRector.php
```diff
@@ -111,13 +112,7 @@ public function refactor(Node $node): ?Node
...
             $originalAttributeValue = $desiredTagValueNode->value->value;
 
-            $methodName = trim($originalAttributeValue, '()');
-
-            $attributeGroup = $this->phpAttributeGroupFactory->createFromClassWithItems(
-                'PHPUnit\Framework\Attributes\DataProvider',
-                [$methodName]
-            );
-            $node->attrGroups[] = $attributeGroup;
+            $node->attrGroups[] = $this->createAttribtueGroup($originalAttributeValue);
```
samsonasik, line 115: typo:

```suggestion
            $node->attrGroups[] = $this->createAttributeGroup($originalAttributeValue);
```

samsonasik:
Lets merge it so we have faster feedback, thank you @staabm

tests/Rector/ClassMethod/DataProviderAnnotationToAttributeRector/Fixture/external-provider-fqcn.php.inc
```diff
@@ -0,0 +1,47 @@
+<?php
```
OskarStark, line 1: This test is the same like below, isn't it?

staabm: There is just a single backslash more :-)

staabm: See https://github.com/rectorphp/rector-phpunit/pull/167#discussion_r1167766572

PR #171, opened 2023-04-16
StaticDataProviderClassMethodRector should only run if `$this` is not used in dataProvider

staabm:
closes https://github.com/rectorphp/rector/issues/7872

### DIFF
--- a/src/Rector/Class_/StaticDataProviderClassMethodRector.php
+++ b/src/Rector/Class_/StaticDataProviderClassMethodRector.php
@@ -5,8 +5,10 @@ declare(strict_types=1);
 namespace Rector\PHPUnit\Rector\Class_;
 
 use PhpParser\Node;
+use PhpParser\Node\Expr\Variable;
 use PhpParser\Node\Stmt\Class_;
 use Rector\Core\Rector\AbstractRector;
+use Rector\NodeNameResolver\NodeNameResolver;
 use Rector\PHPUnit\NodeAnalyzer\TestsNodeAnalyzer;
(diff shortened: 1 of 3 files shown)


src/Rector/Class_/StaticDataProviderClassMethodRector.php
```diff
@@ -108,4 +110,20 @@ public function refactor(Node $node): ?Node
...
+        }
+
+        $variables = $this->betterNodeFinder->findInstanceOf($classMethod, Variable::class);
+        foreach ($variables as $variable) {
+            if ($this->nodeNameResolver->isName($variable, 'this')) {
+                return true;
+            }
+        }
+
+        return false;
```
samsonasik, line 127: This can be changed to `findFirst`

```php
return (bool) $this->betterNodeFinder->findFirst(
            $classMethod->stmts,
            fn (Node $node): bool => $node instanceof Variable && $this->nodeNameResolver->isName($node, 'this')
        );
```

staabm: I don't think it makes it easier to read, but see yourself:

https://github.com/rectorphp/rector-phpunit/pull/171/commits/e5bec06bd9e5235a3409299f6465087503beae8a

samsonasik: `findFirst()` is faster as it stop on found node.

samsonasik (approved):

samsonasik:
Thank you @staabm

PR #194, opened 2023-06-20
[PreferPHPUnitThisCallRector] Make sure to skip static funct fixes #8008

JohJohan:
fixes [#8008](https://github.com/rectorphp/rector/issues/8008)

### DIFF
--- a/src/Rector/Class_/PreferPHPUnitThisCallRector.php
+++ b/src/Rector/Class_/PreferPHPUnitThisCallRector.php
@@ -5,9 +5,11 @@ declare(strict_types=1);
 namespace Rector\PHPUnit\Rector\Class_;
 
 use PhpParser\Node;
+use PhpParser\Node\Expr\Closure;
 use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Expr\StaticCall;
 use PhpParser\Node\Stmt\Class_;
+use PhpParser\Node\Stmt\ClassMethod;
 use PHPStan\Type\ObjectType;
 use Rector\Core\Rector\AbstractRector;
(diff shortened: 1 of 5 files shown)


samsonasik (approved):

src/Rector/Class_/PreferPHPUnitThisCallRector.php
```diff
@@ -73,8 +75,10 @@ public function refactor(Node $node): ?Node
         }
 
         $hasChanged = false;
-        $this->traverseNodesWithCallable($node, function (Node $node) use (&$hasChanged): ?MethodCall {
-            if (! $node instanceof StaticCall) {
+        $isStatic = false;
+        $this->traverseNodesWithCallable($node, function (Node $node) use (&$hasChanged, &$isStatic): ?MethodCall {
+            $isStatic = ($isStatic) || ($node instanceof ClassMethod && $node->isStatic()) || ($node instanceof Closure && $node->static);
```
samsonasik, line 80: oh, `$isStatic` must always be resetted on next visit next method, the ($isStatic) first condition must be removed. The NodeTraverser::DONT_TRAVERSE_CURRENT_AND_CHILDREN should actually can be used when visit ClassMethod or Closure with isStatic condition = true. Could you create another fixture with multiple test, one with static, another one without static method?

JohJohan: What kind of extra test fixture would you like to see we already have a replace fixture

samsonasik: should be something like this:

```php
    public static function testMe()
    {
        self::assertSame(1, 1);
    }

    public function testMe2()
    {
        self::assertSame(1, 1);
    }
```

The `testMe2` should use `$this->assertSame()`

PR #196, opened 2023-06-22
Rename class fixture

samsonasik:
Avoid random error test like in https://github.com/rectorphp/rector-src/actions/runs/5344729318/jobs/9689468923?pr=4315#step:7:26


samsonasik:
The original test itself is invalid, need to be fixed:

```
     /**
      * @dataProvider provideDataSomething()
      */
+    /**
+     * @dataProvider provideDataSomething()
+     */
```

samsonasik:
Fixed 🎉

samsonasik:
This is really strange, local and CI is different

tests/Rector/Class_/YieldDataProviderRector/Fixture/should_not_remove_comments.php.inc
```diff
@@ -38,17 +40,18 @@ use PHPUnit\Framework\TestCase;
...
-    public function test(string $filePath): void
+    public function testSomething(string $filePath): void
     {
     }
 
-    public function provideData(): \Iterator
+    public function provideDataSomething(): \Iterator
     {
+        // Let's pretend that this comment is extremely important and meaningful.
+        // It should not be removed by Rector.
```
samsonasik, line 54: @TomasVotruba this need to be included in "after" part to demonstrate the fixture goal

samsonasik:
@TomasVotruba fixed ;), this also fix the mirror comment before yield ;)

TomasVotruba:
Thanks for the fix ;)

PR #203, opened 2023-06-24
Refactor `AssertTrueFalseToSpecificMethodRector` to create new object when needed

ghostwriter:
instead of creating all objects and storing them in memory via the `__construct` method.


TomasVotruba:
Thank you :+1:

PR #235, opened 2023-07-17
Create failing test-case skip_no_instanceof_testcase.php.inc

toastedghost:
PR for https://github.com/rectorphp/rector/issues/8070

### DIFF
--- a/rules-tests/PHPUnit60/Rector/MethodCall/GetMockBuilderGetMockToCreateMockRector/Fixture/skip_no_instanceof_testcase.php.inc
+++ b/rules-tests/PHPUnit60/Rector/MethodCall/GetMockBuilderGetMockToCreateMockRector/Fixture/skip_no_instanceof_testcase.php.inc
@@ -8,14 +8,14 @@ class MockFactory
     {
     }
 
-    public function create(string $className): \PHPUnit\Framework\MockObject
+    public function create(string $className): \PHPUnit\Framework\MockObject\MockObject
     {
         return $this->getMockBuilder($className)->getMock();
     }
 
-    protected function getMockBuilder(string $className): \PHPUnit\Framework\MockBuilder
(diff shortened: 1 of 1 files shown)


rules-tests/PHPUnit60/Rector/MethodCall/GetMockBuilderGetMockToCreateMockRector/Fixture/skip_no_instanceof_testcase.php.inc
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace Rector\Tests\PHPUnit\PHPUnit60\Rector\MethodCall\GetMockBuilderGetMockToCreateMockRector\Fixture;
+
+class MockFactory
```
samsonasik, line 5: CI green, class may need to be abstract to make create failing test case? 

https://getrector.com/demo/0d3fffb0-7c28-4e37-a482-5de798e6c195

toastedghost: That should not make any difference. I had noticed the error when classes that extended the abstract class and implemented the create method there were changed.

Before the run of rector:

```
public function create(): MockObject
{
        return $this->getMockBuilder($this->className)->getMock();
}
```

after the run of rector:

```
public function create(): MockObject
{
        return $this->createMock($this->className);
}
```

even though the class does not implement a createMock method, nor does it inherit from the PHPUnit TestCase.

I have created a new demo with this class here, where it is refactored without implementing the createMock method nor extending some other class:  https://getrector.com/demo/1245c503-6b96-4b05-a505-694103808802

samsonasik: that's strange, class target seems not exists or require stub

PR #241, opened 2023-08-09
[DX] Remove appendArgs() usage on rector-phpunit rules

samsonasik:
Use native `array_merge()` instead, after all removed, we can remove the `appendArgs()` method from `AbstractRector`.


TomasVotruba:
:+1:

PR #244, opened 2023-08-18
[PhpUnit90] Return null on no change on ExplicitPhpErrorApiRector

samsonasik:
The applied change return null already in line 98 under loop

https://github.com/rectorphp/rector-phpunit/blob/3b48901693f0d960790af65f73b36b36436badd0/rules/PHPUnit90/Rector/MethodCall/ExplicitPhpErrorApiRector.php#L95-L98

### DIFF
--- a/rules/PHPUnit90/Rector/MethodCall/ExplicitPhpErrorApiRector.php
+++ b/rules/PHPUnit90/Rector/MethodCall/ExplicitPhpErrorApiRector.php
@@ -86,7 +86,7 @@ CODE_SAMPLE
     /**
      * @param MethodCall|StaticCall $node
      */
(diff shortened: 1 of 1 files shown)


samsonasik:
see why it should return null, it show on the list even no change based on requirement

https://getrector.com/demo/26ffe9ca-8ee4-4f0a-a107-774128315061

samsonasik:
PHPStan notice seems unrelated as it happen on main branch as well

https://github.com/rectorphp/rector-phpunit/actions/runs/5902201371/job/16009773664#step:5:19

TomasVotruba:
Looks good, let's merge it :+1: 

I'll finish up the fix on main soon so we'll be green :)

PR #255, opened 2023-09-14
Handle non static methods in data providers

themark147:


rules-tests/PHPUnit100/Rector/Class_/StaticDataProviderClassMethodRector/Fixture/handle_non_static_methods_in_data_provider.php.inc
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * @dataProvider provideData
+     */
+    public function test()
+    {
+    }
+
+    public static function provideData()
+    {
+        yield [static::internalMethod()];
```
samsonasik, line 46: use self instead of static as it has more wider support on multiple php version 

```suggestion
        yield [self::internalMethod()];
```

samsonasik:
Could you try provide a patch as well, it should be something like this:

```php
            $this->traverseNodesWithCallable($dataProviderClassMethod, function (Node $subNode) {
                if ($subNode instanceof Class_ || $subNode instanceof Node\Stmt\Function_ || $subNode instanceof Node\Expr\Closure) {
                    return \PhpParser\NodeTraverser::DONT_TRAVERSE_CURRENT_AND_CHILDREN;
                }

                if (! $subNode instanceof Node\Expr\MethodCall) {
                    return null;
                }

                if ($subNode->isFirstClassCallable()) {
                    return null;
                }

                if (! $this->isName($subNode->var, 'this')) {
                    return null;
                }

                if (! $subNode->name instanceof Node\Identifier) {
                    return null;
                }

                // use self:: instead of static::
                return $this->nodeFactory->createStaticCall('self', $subNode->name->toString(), $subNode->getArgs());
            });
```

before set:

```php
$hasChanged = true;
```

on line 108.

themark147:
@samsonasik Hey, I already implemented a patch what u provide ... hope its fine :))

samsonasik:
Could you CI error :) ? You can reproduce locally for that, first, install the dependencies:

```
export COMPOSER_ROOT_VERSION=dev-main && composer install
```

Then, run:

```
vendor/bin/phpunit
```

then patch it, after that, run:

```bash
```

then, verify if there is phpstan error:

```
vendor/bin/phpstan
```

if the error is code complexity, move the code to separate private method, then run once more:

```
```

 Thank you.

themark147:
@samsonasik I already tried use that patch to refactor real project and it refactor well just called methods still remains NON static

samsonasik:
this currently skipped on latest main, so I am closing it.

PR #258, opened 2023-09-18
Fix/158 Add Min php version for WithConsecutiveRector rule

eerison:
Issue #158

### DIFF
--- a/src/Rector/StmtsAwareInterface/WithConsecutiveRector.php
+++ b/src/Rector/StmtsAwareInterface/WithConsecutiveRector.php
@@ -18,14 +18,16 @@ use PhpParser\Node\Scalar\LNumber;
 use PhpParser\Node\Stmt\Expression;
 use PhpParser\Node\Stmt\Return_;
 use Rector\Core\Rector\AbstractRector;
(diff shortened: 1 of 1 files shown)


eerison:
Just for curiosity, how Could I reproduce this error locally?
what command is executed in this step?

samsonasik:
You can test rector/rector:dev-main in next 5 minutes

TomasVotruba:
We use shared repository for the CI. These are the exact commands:

* https://github.com/rectorphp/reusable-workflows/blob/main/.github/workflows/rector.yaml#L28-L29
* https://github.com/rectorphp/reusable-workflows/blob/main/.github/workflows/code_analysis.yaml#L18-L23

samsonasik:
To run composer install, you need to define:

```
export COMPOSER_ROOT_VERSION=dev-main && composer install

**Comment by eerison** [comment]
> To run composer install, you need to define:
> 
> ```
> export COMPOSER_ROOT_VERSION=dev-main && composer install
> ```

hmmm ok,  I will try that :)

**Comment by samsonasik** [comment]
Yes, once done, you can run:

```

PR #263, opened 2023-09-26
Replace decimal delta value for PHP_FLOAT_EPSILON

eerison:
In php 7.2 was introduced [PHP_FLOAT_EPSILON](https://www.php.net/manual/en/reserved.constants.php#constant.php-float-epsilon) and I could use this as a small representation for decimal delta.

https://github.com/sebastianbergmann/phpunit/pull/4874
https://github.com/sebastianbergmann/phpunit/issues/3159#issuecomment-1022240982.

and I tested with diffident values and it worked as expected.


TomasVotruba:
This looks great, thanks for investigating :+1:

PR #264, opened 2023-09-27
PHPUnit10 Attributes

avgeeklucky:
Changed refactors done in `CoversAnnotationWithValueToAttributeRector.php` so that
- [CoversClass](https://github.com/sebastianbergmann/phpunit/blob/main/src/Framework/Attributes/CoversClass.php), [CoversFunction](https://github.com/sebastianbergmann/phpunit/blob/main/src/Framework/Attributes/CoversFunction.php)  attributes are added on class level (in accordance with their respective `#[Attribute ... ]` values). 
- `@covers` annotations are removed from method level
- Duplicate covers attributes are unified
- `@coversDefault` is converted to a `CoversClass` attribute

### DIFF
--- a/docs/rector_rules_overview.md
+++ b/docs/rector_rules_overview.md
@@ -378,12 +378,12 @@ Change covers annotations with value to attribute
 - * @covers SomeClass
 - */
 +#[CoversClass(SomeClass::class)]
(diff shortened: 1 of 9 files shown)


TomasVotruba:
Thank you :clap:

PR #266, opened 2023-10-07
Remove codeCoverageIgnore from annotations to attributes set

marmichalski:
This is no longer deprecated as of 10.4 https://github.com/sebastianbergmann/phpunit/blob/10.4/ChangeLog-10.4.md#changed.

### DIFF
--- a/config/sets/annotations-to-attributes.php
+++ b/config/sets/annotations-to-attributes.php
@@ -88,7 +88,6 @@ return static function (RectorConfig $rectorConfig): void {
         new AnnotationToAttribute('afterClass', 'PHPUnit\Framework\Attributes\AfterClass'),
         new AnnotationToAttribute('before', 'PHPUnit\Framework\Attributes\Before'),
         new AnnotationToAttribute('beforeClass', 'PHPUnit\Framework\Attributes\BeforeClass'),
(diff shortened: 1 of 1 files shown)


TomasVotruba:
What is a benefit over native attribute?
The idea of upgrade is not to avoid the deprecated code, but rather use the best PHP version as soon as possible :)

marmichalski:
The benefit is that the attribute has now been deprecated :)

https://github.com/sebastianbergmann/phpunit/blob/10.4/src/Framework/Attributes/CodeCoverageIgnore.php#L19

So rector with phpunit 10.4 wants to change non deprecated annotation to deprecated attribute.

TomasVotruba:
I see. I think this will need better migration once PHPUnit 11 lands. Adding and removing annotations/attributes back and forth would break the upgrade flow.

Let's wait what PHPUnit 11 brings :)

marmichalski:
Sure, but currently rector conflicts with phpstan inspections, as I don't think there is any easy way to remove that single conversion in consuming project? Other than overriding the service probably?

TomasVotruba:
I'd ignore this in PHPStan, as it's a single line easy solution.

Issue #267, opened 2023-10-09
[PHPUnit 10] Change setMethods() to onlyMethods()


TomasVotruba:
See https://stackoverflow.com/questions/65075204/method-setmethods-is-deprecated-when-try-to-write-a-php-unit-test

eerison:
yep it was deprecated in 9.x and already dropped in 10.x

https://statics.teams.cdn.office.net/evergreen-assets/safelinks/1/atp-safelinks.html

eerison:
I'm on this issue ;)

eerison:
Hey @TomasVotruba 

don't you think we could add this into `PhpUnit90` rules insteadof `PhpUnit10`?

TomasVotruba:
Hey, we usually add the rule when the change will affect users. When people upgrade to PHPUnit 9.0 and this change is not there yet, it would break.

So PHPUnit 10.0 is safer

eerison:
It does not work for setMethods :'(

https://github.com/rectorphp/rector-phpunit/commit/6d9316d7c1c4ba206fe7904adfeb03521f26cc2b

eerison:
it works just in the first case


```php
        //first
        $this->setMethods();

        //second
        $contractMock = $this->getMockBuilder(Contract::class)
            ->disableOriginalConstructor()
            ->setMethods(['getCancelledAt', 'terminate'])
            ->getMock();
```

samsonasik:
I am on mobile, `disableOriginalConstructor()` seems return different instance?

eerison:
> I am on mobile, `disableOriginalConstructor()` seems return different instance?

I see the issue, I will provide a PR for this

samsonasik:
Resolved at https://github.com/rectorphp/rector-phpunit/pull/280

eerison:
I was wondering why this issue wasn't close in the first PR, maybe you have the feling that it wasn't fixed 😄 

Just to confirm that it worked <3

PR #277, opened 2023-10-26
Feature/267 Add rule to replace setMethods to onlyMethods

eerison:
Issue #267

### DIFF
--- a/config/sets/phpunit100.php
+++ b/config/sets/phpunit100.php
@@ -62,5 +62,8 @@ return static function (RectorConfig $rectorConfig): void {
 
         // https://github.com/sebastianbergmann/phpunit/issues/4063
         new MethodCallRename('PHPUnit\Framework\Assert', 'assertNotIsReadable', 'assertIsNotReadable'),
+
+        // https://github.com/sebastianbergmann/phpunit/pull/3687
+        new MethodCallRename('PHPUnit\Framework\Assert', 'setMethods', 'onlyMethods'),
     ]);
 };


eerison:
Wait for this PR https://github.com/rectorphp/rector-phpunit/pull/278

rules/PHPUnit90/Rector/MethodCall/UseOnlyMethodRector.php
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * @param MethodCall $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if (! $this->testsNodeAnalyzer->isPHPUnitMethodCallNames($node, ['setMethods'])) {
+            return null;
+        }
+
```
samsonasik, line 66:
```suggestion

        if ($node->isFirstClassCallable()) {
            return null;
        }
```

rules/PHPUnit90/Rector/MethodCall/UseOnlyMethodRector.php
```diff
@@ -0,0 +1,74 @@
...
+        $args = $node->getArgs();
+
```
samsonasik, line 71:
```suggestion
        if (! isset($node->args[0])) {
            return null;
        }
        
```

rules/PHPUnit90/Rector/MethodCall/UseOnlyMethodRector.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\PHPUnit\PHPUnit90\Rector\MethodCall;
```
samsonasik, line 5: is it phpunit 9 or phpunit 10? If phpunit 10, then it needs in `PHPUnit100` namespace imo:

it also may need to be added to the phpunit 10 setlist

https://github.com/rectorphp/rector-phpunit/blob/main/config/sets/phpunit100.php

eerison: it was deprecated phpuni9 and removed phpunit10 in this case do we apply for versions that we are removing?

samsonasik: I see, but it seems still need to be registered to `config/sets/phpunit100.php`

eerison: then should I move to PhpUnit100 folder and register in `config/sets/phpunit100.php`?

samsonasik: I think using `PHPUnit` directory is better, but probably @TomasVotruba has opinion?

eerison: Hey @samsonasik 

I guess I could just use 

it 😄 
```php
new MethodCallRename('PHPUnit\Framework\Assert', 'setMethods', 'onlyMethods'),
```

eerison: > https://github.com/rectorphp/rector-phpunit/blob/main/config/sets/phpunit100.php

like this https://github.com/rectorphp/rector-phpunit/blob/main/config/sets/phpunit100.php#L58 😄

samsonasik: yeah, that seems good 👍

samsonasik (approved):

TomasVotruba:
Thank you

PR #278, opened 2023-10-26
New fix after run rector

eerison:
after I update packages locally new rules was applied.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,6 +44,7 @@
         "phpstan": "vendor/bin/phpstan analyse --ansi --error-format symplify",
         "check-cs": "vendor/bin/ecs check --ansi",
         "fix-cs": "vendor/bin/ecs check --fix --ansi",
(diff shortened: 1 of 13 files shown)


composer.json
```diff
@@ -44,6 +44,8 @@
         "phpstan": "vendor/bin/phpstan analyse --ansi --error-format symplify",
         "check-cs": "vendor/bin/ecs check --ansi",
         "fix-cs": "vendor/bin/ecs check --fix --ansi",
+        "check-rector": "vendor/bin/rector process --dry-run --ansi",
+        "fix-rector": "vendor/bin/rector process --ansi",
```
samsonasik, line 48: to sync with what applied in rector-src:

```suggestion
        "rector": "vendor/bin/rector process --ansi",
```

eerison: Done ;)

samsonasik:
Thank you @eerison

PR #282, opened 2023-10-28
Removed @codeCoverageIgnore to attribute conversion

vudaltsov:
See https://github.com/sebastianbergmann/phpunit/issues/5513.


TomasVotruba:
This change should be in a new PHPUnit 11 set

vudaltsov:
> This change should be in a new PHPUnit 11 set

Even though it's deprecated in 10.4? Support for phpDoc annotations was never dropped, so it's sage to use them in any 10.x version.

TomasVotruba:
Yes, as this would not be valid on PHPUnit 10.0.
We move the changes to next major/breaking release to be sure the features all work 100 % on the major.0 release

vudaltsov:
> Yes, as this would not be valid on PHPUnit 10.0.

I disagree. In fact it **is** valid for PHPUnit `10.0`, `10.1`, `10.2` and `10.3`. So if you have "phpunit/phpunit": "10.0.0" and a class with `@codeCoverageIgnore` and without attribute, it will be ignored in coverage report. No warning will be triggered.

So it is safe to merge PR as is.

TomasVotruba:
This our convention for all sets we use.

PR #283, opened 2023-11-01
Feature/replace testWith annotation to dataProviders

eerison:
Issue #281

### DIFF
--- a/config/sets/phpunit-code-quality.php
+++ b/config/sets/phpunit-code-quality.php
@@ -6,6 +6,7 @@ use Rector\Config\RectorConfig;
 use Rector\PHPUnit\CodeQuality\Rector\Class_\AddSeeTestAnnotationRector;
 use Rector\PHPUnit\CodeQuality\Rector\Class_\ConstructClassMethodToSetUpTestCaseRector;
 use Rector\PHPUnit\CodeQuality\Rector\Class_\PreferPHPUnitThisCallRector;
+use Rector\PHPUnit\CodeQuality\Rector\Class_\TestWithToDataProviderRector;
 use Rector\PHPUnit\CodeQuality\Rector\Class_\YieldDataProviderRector;
 use Rector\PHPUnit\CodeQuality\Rector\ClassMethod\RemoveEmptyTestMethodRector;
 use Rector\PHPUnit\CodeQuality\Rector\ClassMethod\ReplaceTestAnnotationWithPrefixedFunctionRector;
@@ -38,6 +39,7 @@ return static function (RectorConfig $rectorConfig): void {
         YieldDataProviderRector::class,
         RemoveEmptyTestMethodRector::class,
         ReplaceTestAnnotationWithPrefixedFunctionRector::class,
+        TestWithToDataProviderRector::class,
 
         // specific asserts
         AssertCompareToSpecificMethodRector::class,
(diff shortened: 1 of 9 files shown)


rules/Transform/Rector/Class_/TestWithToDataProviderRector.php
```diff
@@ -0,0 +1,180 @@
...
+
+        if(is_string($value)) {
+            return new Node\Scalar\String_($value);
+        }
+
+        if(is_integer($value)) {
+            return new Node\Scalar\LNumber($value);
+        }
+
+        return new Node\Scalar\DNumber($value);
```
eerison, line 178: Hello guys

I need some support here :(, I do not know if it is the best/only way to parse those values.

But to make this works I need to use `ArrayItem` and ArrayItem requires a object `Expr`.


But for some cases like `bool` and `null` values, I do not have a Scalar class for this (well I did not find).

it is just missing make those tests pass, you can check into `feature3.php.inc`.

cc @TomasVotruba or @samsonasik

rules-tests/Transform/Rector/Class_/TestWithToDataProviderRector/Fixture/fixture2.php.inc
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Rector\PHPUnit\Tests\Transform\Rector\Class_\TestWithToDataProviderRector\Fixture;
+
+use PHPUnit\Framework\TestCase;
+
+class MyTest1 extends TestCase
+{
+    public function dataProvider()
+    {
+        return [[[['foo', 'bar', 1], ['foo', 'bar', 2]]], [[['foo', 1]]]];
```
eerison, line 31: Reminder for me

it is missing add key values, like `fieldA => foo`,

rules/Transform/Rector/Class_/TestWithToDataProviderRector.php
```diff
@@ -0,0 +1,209 @@
+<?php
```
eerison, line 1: Please also check if the files are placed in the correct folders path :)

Note: in case it's, then I need to provide other PR moving this file to rules:
https://github.com/rectorphp/rector-phpunit/blob/main/src/Transform/AssertEqualsOrAssertSameFloatParameterToSpecificMethodsTypeRector.php, when I was doing this feature, I realised that is in the wrong place :/

TomasVotruba: Ideally put it here, so we use existing set and avoid too detailed granulation.

https://github.com/rectorphp/rector-phpunit/tree/main/rules/CodeQuality/Rector

Once you move it, it's ready to merge

eerison: Do you mean this new rule ([TestWithToDataProviderRector.php](https://github.com/rectorphp/rector-phpunit/pull/283/files#diff-6d31e06ec7631468c3201fef89d14a831a0c792c782389beff876353cba3ad94)) or this other one that is already merged ([AssertEqualsOrAssertSameFloatParameterToSpecificMethodsTypeRector.php](https://github.com/rectorphp/rector-phpunit/blob/main/src/Transform/AssertEqualsOrAssertSameFloatParameterToSpecificMethodsTypeRector.php))?

or do you mean both of them?

TomasVotruba: This one in this PR - `TestWithToDataProviderRector`

eerison: Hello @TomasVotruba Dobré ráno :)

how are you, I hope you have had a great start of week :)

I just moved the code to CodeQuality and I also added it in `config/sets/phpunit-code-quality.php`, I hope everything as expected 😄 

Thanks

TomasVotruba:
Thank you, looks good to go :+1:

PR #285, opened 2023-11-06
Move AssertEqualsOrAssertSameFloatParameterToSpecificMethodsTypeRector to CodeQuality

eerison:
@TomasVotruba I guess we should also move this rule for CodeQuality, WDYT?


TomasVotruba:
Yes, that's perfect! Thanks :+1:

PR #286, opened 2023-11-06
Support testWith multiline

eerison:
Hello, I did not realise that is possible to use `testWith multiline`.

I adjust the code to work with this

Close #281


TomasVotruba:
Awesome, this was tricky one in PHP 8 attributes migration too, so I feel you :)

Thanks for handling it :+1:

PR #291, opened 2023-11-08
[CodeQuality] Skip inline comment on DataProviderArrayItemsNewLinedRector

samsonasik:
array with inline comment should be skipped instead of re-printed, it currently produce invalid change:

```diff
     public function provideData(): array
     {
         return [
-            ['content123', 11], // a comment inline
+            ['content123', 11],
```


samsonasik:
Fixed with verify next array item start line is equal with current array item line https://github.com/rectorphp/rector-phpunit/pull/291/commits/0aade184eaaf620717cfdd3079984d3cdcfc9b6a

samsonasik:

PR #293, opened 2023-11-10
setMethods call has been removed in PHPUnit 10

bbrala:
Deprecated in PPHPUnit 8.3 see: https://github.com/sebastianbergmann/phpunit/commit/d618fa3fda437421264dcfa1413a474f306f79c4

Fixes #292

PR #294, opened 2023-11-11
RemoveSetMethodsMethodCallRector: move tests and classes to new location

bbrala:
Move RemoveSetMethodsMethodCallRector and its tests to correct namespace.

As per: https://github.com/rectorphp/rector-phpunit/pull/293#issuecomment-1806286903

### DIFF
--- a/config/sets/phpunit100.php
+++ b/config/sets/phpunit100.php
@@ -3,10 +3,10 @@
 declare(strict_types=1);
 
 use Rector\Config\RectorConfig;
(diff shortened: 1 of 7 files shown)


bbrala:
Failure is because the branch is not in main repo i think?

samsonasik:
You can run:

```
vendor/bin/rector && vendor/bin/ecs check --fix
```

to make CI green

bbrala:
Ty will do

samsonasik:
Thank you @bbrala let's give it a try :+1:

samsonasik:
it seems working now, see:

- https://github.com/rectorphp/rector-src/pull/5242

PR #300, opened 2023-12-26
[PHPUnit 10] Handle DataProvider attribute not included on use PHPUnitSetList::PHPUNIT_100

samsonasik:
Fixes https://github.com/rectorphp/rector/issues/8360


samsonasik:
@rpkamp the issue is due to the class not extending `\PHPUnit\Framework\TestCase`, change to extends it solve it, tested with this PR.

@TomasVotruba I am merging it as future proven for test.

TomasVotruba:
Thank you :+1:

PR #301, opened 2023-12-29
[PHPUnit100] Adds the PublicDataProviderClassMethodRector rule

peterfox:
# Changes

* Adds a new rule for making data provider methods public.
* Adds a new set of tests for the rule.
* Updates docs
* Adds the rule to the PHPUnit100 config.

# Why

PHPUnit 10 expects data providers to be both Public and Static whereas before both were not a requirement. There was already a rule for static data providers but protected/private scenario hasn't been covered. For ease of use I've created a new rule rather than changing the static rule to cover both scenarios.

### DIFF
--- a/config/sets/phpunit100.php
+++ b/config/sets/phpunit100.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 use Rector\Config\RectorConfig;
 use Rector\PHPUnit\PHPUnit100\Rector\Class_\AddProphecyTraitRector;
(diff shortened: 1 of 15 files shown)


TomasVotruba:
Seems good to go :+1:  Just one detail from rector CI workflow

peterfox:
Thanks @TomasVotruba, I've refactored and will hopefully have fixed it

TomasVotruba:
Thanks :+1:  Let's see.

If you want to avoid waiting for CI, you can do the same check locally:

```
vendor/bin/rector
```

TomasVotruba:
All good, thank you :clap:

PR #304, opened 2024-01-03
Remove `codeCoverageIgnore` to annotation to attribute conversion

### LINKED ISSUE #303: CodeCoverageIgnore attribute is deprecated and should not be used
As described in https://github.com/sebastianbergmann/phpunit/issues/5235 and https://github.com/sebastianbergmann/phpunit/issues/5513 the `CodeCoverageIgnore` attribute was a mistake and is now deprecated, therefore Rector should not try to replace annotation with the attribute.

https://github.com/rectorphp/rector-phpunit/blob/2095737213ec328db197774a5d073e962c14174c/config/sets/annotations-to-attributes.php#L91

martinssipenko:
As described in https://github.com/sebastianbergmann/phpunit/issues/5235 and https://github.com/sebastianbergmann/phpunit/issues/5513 the `CodeCoverageIgnore` attribute was a mistake and is now deprecated, therefore Rector should not try to replace annotation with the attribute.

Fixes #303

### DIFF
--- a/config/sets/annotations-to-attributes.php
+++ b/config/sets/annotations-to-attributes.php
@@ -88,7 +88,6 @@ return static function (RectorConfig $rectorConfig): void {
         new AnnotationToAttribute('afterClass', 'PHPUnit\Framework\Attributes\AfterClass'),
         new AnnotationToAttribute('before', 'PHPUnit\Framework\Attributes\Before'),
         new AnnotationToAttribute('beforeClass', 'PHPUnit\Framework\Attributes\BeforeClass'),
-        new AnnotationToAttribute('codeCoverageIgnore', 'PHPUnit\Framework\Attributes\CodeCoverageIgnore'),
         new AnnotationToAttribute('coversNothing', 'PHPUnit\Framework\Attributes\CoversNothing'),
         new AnnotationToAttribute('doesNotPerformAssertions', 'PHPUnit\Framework\Attributes\DoesNotPerformAssertions'),
         new AnnotationToAttribute('large', 'PHPUnit\Framework\Attributes\Large'),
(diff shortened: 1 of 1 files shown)


config/sets/annotations-to-attributes.php
```diff
@@ -88,7 +88,6 @@
         new AnnotationToAttribute('afterClass', 'PHPUnit\Framework\Attributes\AfterClass'),
         new AnnotationToAttribute('before', 'PHPUnit\Framework\Attributes\Before'),
         new AnnotationToAttribute('beforeClass', 'PHPUnit\Framework\Attributes\BeforeClass'),
-        new AnnotationToAttribute('codeCoverageIgnore', 'PHPUnit\Framework\Attributes\CodeCoverageIgnore'),
```
TomasVotruba, line 91: Could you add links to the 2 issues and not about code coverage being deprecated? 
Just so someone don't re-add it here in near future.

martinssipenko: Add links where? In code?

samsonasik: @martinssipenko probably comment the line instead of remove and add `@see` link doc

TomasVotruba: Here in this line

samsonasik:
@martinssipenko could you update workflow file to require php 8.2 as it require rector-src? Thank you

samsonasik:
@martinssipenko latest main already use php 8.2, could you rebase?

martinssipenko:
@samsonasik @TomasVotruba updated

TomasVotruba:
Awesome, thank you :+1:

PR #306, opened 2024-01-09
WithConsecutiveRector was applied only to dynamic invocations count

voronovvladimir:
PR is intended to give ability to refactor withConsecutive when static invocation checker applied
 
According to PHPUnit [there is no difference between static and non-static assertions](https://docs.phpunit.de/en/10.5/assertions.html#static-vs-non-static-usage-of-assertion-methods) but it sounds like invocation count checkers have the same ability and there is no difference between self::exactly and $this->exactly


TomasVotruba:
Looks good, thank you :+1:

PR #315, opened 2024-03-05
[PHPUnit][CodeQuality] Skip no arg on AssertEmptyNullableObjectToAssertInstanceofRector

samsonasik:
Fixes https://github.com/rectorphp/rector/issues/8527 for whatever reason to not have arg on the first place, eg: different object caller.


samsonasik:
All checks have passed 🎉 @TomasVotruba I am merging it;)

Issue #318, opened 2024-03-23
Add new rector rule to add `#[CoversClass()]` to test classes


maks-rafalko:
Desired behavior:

```diff
+ #[CoversClass(SomeService::class)]
class SomeServiceTest extends TestCase
{
}
```

Actually, I've already implemented it as a custom rule as we need it in Infection. [It works great](https://github.com/infection/infection/pull/1945/files) and covered by tests

https://github.com/infection/infection/commit/03da55739a01e38f2ddf3dbef9f0f7e6adf53f42#diff-8a5ee371c08d2d066c7dfe04dea926a679e3ff5fb35ff93a228aa2a5c46e3699R65

But I would like to integrate it into `rector/rector-phpunit` itself, as I'm pretty sure it will be useful for other teams that would like to *automatically* add `#[CoversClass()]` to their test classes and *automatically* guess source file used as a value of this attribute. This is pretty convenient, and existing tools, including PHP-CS-Fixer, doesn't help with it.

But before creating a PR, I would like to know if it's going to be accepted to not waste time.

Also, this issue will popup in a search so even if it's not going to be accepted, one will have a link to a custom implementation.

TomasVotruba:
Looks good :+1: 

Please send the PR :)

maks-rafalko:
Ready for review: https://github.com/rectorphp/rector-phpunit/pull/319

PR #319, opened 2024-03-24
Add rector rule to automatically add `#[CoversClass]` attribute to test files

maks-rafalko:
Implements https://github.com/rectorphp/rector-phpunit/issues/318

# Rationale

On PHPUnit side, we can configure our tests in the way that every Test File has to have `#[CoversClass(...)]` attributes, otherwise tests will be marked as `R`isky and PHPUnit will exit with non-zero error code.

This can be done using 

- [`requireCoverageMetadata`](https://docs.phpunit.de/en/11.0/configuration.html#the-requirecoveragemetadata-attribute) configuration attribute (previously `forceCoversAnnotation="true"`)

Why is it needed?

- because it's a good practice (e.g. read https://localheinz.com/articles/2023/02/22/documenting-the-system-under-test-in-phpunit/)
- and for us, [Infection](https://github.com/infection/infection) maintainers, it helps to easier introduce new [`--map-source-class-to-test`](https://infection.github.io/guide/command-line-options.html#map-source-class-to-test) option to **dramatically** speed up Mutation Testing (MT) for projects

Real example with PHP-CS-Fixer: https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/7874#issuecomment-2010894591

We managed to decrease the MT execution from `2m 27s` to `8s` (!!!). And this was possible because in PHP-CS-Fixer their tests has 1-to-1 relation between test file and source file, thanks to `#[CoversClass(...)]` annotation.

But, I wasn't able to find an _automatic_ way of how we can:

- update existing project and introduce this attribute to all the tests files (at Infection, without this feature I would need to [manually update ~400 files](https://github.com/infection/infection/pull/1945))
- force this attribute on each PR when new test file is created.

So, I ended up creating this PR that will help other developers to achieve the same goals.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,4 +1,4 @@
-    parameters:
+parameters:
     level: 8
 
     paths:
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,4 +1,4 @@
-parameters:
+    parameters:
     level: 8
 
     paths:
@@ -27,9 +27,3 @@ parameters:
         - '#Access to an undefined property Rector\\Contract\\PhpParser\\Node\\StmtsAwareInterface\:\:\$stmts#'
 
         - '#PhpParser\\Node\\Stmt\\Expression is not generic#'
-
-        # false positive
-        -
-            path: 'rules/CodeQuality/Rector/Class_/AddCoversClassAttributeRector.php'
-            message: '#Parameter \#2 \$callback of function array_filter expects \(callable\(string\|null\)\: bool\)\|null, Closure\(string\)\: bool given#'
-            count: 1
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -27,3 +27,9 @@ parameters:
         - '#Access to an undefined property Rector\\Contract\\PhpParser\\Node\\StmtsAwareInterface\:\:\$stmts#'
 
         - '#PhpParser\\Node\\Stmt\\Expression is not generic#'
+
+        # false positive
+        -
+            path: 'rules/CodeQuality/Rector/Class_/AddCoversClassAttributeRector.php'
+            message: '#Parameter \#2 \$callback of function array_filter expects \(callable\(string\|null\)\: bool\)\|null, Closure\(string\)\: bool given#'
+            count: 1
--- a/rules-tests/CodeQuality/Rector/Class_/AddCoversClassAttributeRector/config/configured_rule.php
+++ b/rules-tests/CodeQuality/Rector/Class_/AddCoversClassAttributeRector/config/configured_rule.php
@@ -1,35 +1,4 @@
 <?php
-/**
- * This code is licensed under the BSD 3-Clause License.
- *
- * Copyright (c) 2017, Maks Rafalko
- * All rights reserved.
- *
- * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions are met:
- *
- * * Redistributions of source code must retain the above copyright notice, this
- *   list of conditions and the following disclaimer.
- *
- * * Redistributions in binary form must reproduce the above copyright notice,
- *   this list of conditions and the following disclaimer in the documentation
- *   and/or other materials provided with the distribution.
- *
- * * Neither the name of the copyright holder nor the names of its
- *   contributors may be used to endorse or promote products derived from
- *   this software without specific prior written permission.
- *
- * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
- * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
- * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
- * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
- * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
- * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
- * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
- * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
- * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
- * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
- */
 
 declare(strict_types=1);
 
--- a/rules-tests/CodeQuality/Rector/Class_/AddCoversClassAttributeRector/config/configured_rule.php
+++ b/rules-tests/CodeQuality/Rector/Class_/AddCoversClassAttributeRector/config/configured_rule.php
@@ -1,0 +1,41 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of the copyright holder nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+declare(strict_types=1);
+
+use Rector\Config\RectorConfig;
+use Rector\PHPUnit\CodeQuality\Rector\Class_\AddCoversClassAttributeRector;
+
+return static function (RectorConfig $rectorConfig): void {
+    $rectorConfig->rule(AddCoversClassAttributeRector::class);
+};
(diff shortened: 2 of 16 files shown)


phpstan.neon
```diff
@@ -27,3 +27,9 @@ parameters:
         - '#Access to an undefined property Rector\\Contract\\PhpParser\\Node\\StmtsAwareInterface\:\:\$stmts#'
 
         - '#PhpParser\\Node\\Stmt\\Expression is not generic#'
+
+        # false positive
+        -
+            path: 'rules/CodeQuality/Rector/Class_/AddCoversClassAttributeRector.php'
+            message: '#Parameter \#2 \$callback of function array_filter expects \(callable\(string\|null\)\: bool\)\|null, Closure\(string\)\: bool given#'
```
maks-rafalko, line 34: I don't get this error and can't understand what's wrong with this code:

```php
$partsWithoutTests = array_filter(
    $classNameParts,
    static fn (string $part): bool => ! in_array(strtolower($part), ['test', 'tests'], true),
);
```

samsonasik: You may can lookup this discussion 

- https://github.com/phpstan/phpstan/discussions/8911

maks-rafalko: ok, interesting. Thank you for the link

maks-rafalko:
```

Classes with a parent/interface - possibly used by type
=======================================================

 * Rector\PHPUnit\CodeQuality\Rector\Class_\AddCoversClassAttributeRector
rules/CodeQuality/Rector/Class_/AddCoversClassAttributeRector.php
```

should I add it to `config/sets/phpunit-code-quality.php` to fix this issue?

By the way, `composer.json` has different scripts to check code quality (phpstan, ecs, etc.). But `class-leak` is no present there, that's why I event didn't know about it as a first-time contributor.

Should it be added too?

https://github.com/rectorphp/rector-phpunit/blob/6845db43ccd69ef990d399ff845a53ad66fa8085/composer.json#L36-L42

VincentLanglet:
If I understand correctly the rule, it only works if the test and the tested class has the same namespace. Am I wrong ?

In Symfony doc (https://symfony.com/doc/current/create_framework/unit_testing.html and https://symfony.com/doc/current/testing.html), the namespace used are more often
- App\Foo, for the class
- App\Tests\Foo, for the test

1) Would it work ?
2) Would it be easy to add an option for the "namespace mapping"

(In my case, I use `App\Foo` and `App\Tests\Unit\Foo` or `App\Tests\Functional\Foo` ; so an option would be great)

maks-rafalko:
> If I understand correctly the rule, it only works if the test and the tested class has the same namespace.

No. The current implementation adds 2 possible variations to find a source class based on the name of the test class (logic is [here](https://github.com/rectorphp/rector-phpunit/pull/319/files#diff-71d17b994a929ea88aa8993bc28aadddef08a0f7f1a12f571bfb49478325165aR122-R141))

If you have `App\Tests\SubNamespace\FooTest` (note `*Test` postfix, it's absent for some reason in your example), the followig candidates will be generated:

- `App\Tests\SubNamespace\Foo` - same namespace, but without `*Test` postfix. This will be useful for those projects who store test classes near source classes. But in real project case FQCNs would be  `App\SubNamespace\Foo` for source class and `App\SubNamespace\FooTest` for test class
- `App\SubNamespace\Foo` - `Tests` subnamespace is removed

If you have namespaces for tests containing `Unit` / `Functional` subnamespaces, which don't present in source class FQCN, then it wouldn't find a source file with the current implementation.

Of course, we can make the logic of resolving source class name more complex, but I would like to first understand if it's really needed and will be accepted, just to not waste time.

---

> 
    App\Foo, for the class
    App\Tests\Foo, for the test

    Would it work ?

yes

> Would it be easy to add an option for the "namespace mapping"

could you please describe how it would work?

VincentLanglet:
> If you have `App\Tests\SubNamespace\FooTest` (note `*Test` postfix, it's absent for some reason in your example)

i meant `App\Tests\Foo\ServiceTest`, it just wrote the namespace, but yes
`App\Tests\SubNamespace\FooTest` is definitly what I had in mind.

> > Would it be easy to add an option for the "namespace mapping"
> 
> could you please describe how it would work?

Not sure what was the best solution, I thought we could have a `$testMapping` option, by default it's
for instance in my personal use case I would have pass something like
```
[
    'App\Test\Unit' => 'App',
    'App\Test\Functional' => 'App',
]
```

But I definetly understand it's a special use case and might not be worth it in this generic rule.

maks-rafalko:
other solution would be to try generate more combinaions, removing commonly used subnamespaces, e.g. for `App\Tests\Unit\Foo` generate:

- `App\Unit\Foo` - remove `Tests`
- `App\Unit\Foo` - remove `Tests` and `Unit` (where both are reserved words

But yeah, let's first listen to maintainers before doing any changes. Probably makes sense to merge the common case and then upgrade resolving logic as a separate PR (if needed).

rules-tests/CodeQuality/Rector/Class_/AddCoversClassAttributeRector/config/configured_rule.php
```diff
@@ -0,0 +1,41 @@
...
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
```
samsonasik, line 32: I think this doc license is copy paste mistake

rules/CodeQuality/Rector/Class_/AddCoversClassAttributeRector.php
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        if (! $this->testsNodeAnalyzer->isInTestClass($node)) {
+            return null;
+        }
+
+        if ($this->phpAttributeAnalyzer->hasPhpAttributes($node, [
+            CoversNothing::class,
+            CoversClass::class,
+            CoversFunction::class,
```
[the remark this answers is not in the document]
maks-rafalko: Done, but could you please explain why on downgrade it's being prefixed?

rules-tests/CodeQuality/Rector/Class_/AddCoversClassAttributeRector/config/configured_rule.php
maks-rafalko: oops, copied from Infection

rules/CodeQuality/Rector/Class_/AddCoversClassAttributeRector.php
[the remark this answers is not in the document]
samsonasik: Rector scoped with php-scoper that included scoped vendor, so namespace is prefixed to avoid conflict

rules/CodeQuality/Rector/Class_/AddCoversClassAttributeRector.php
```diff
@@ -0,0 +1,164 @@
...
+
+        $node->attrGroups = array_merge($node->attrGroups, [$coversAttributeGroup]);
+
+        return $node;
+    }
+
+    /**
+     * @return string[]
+     */
+    public function resolveSourceClassNames(string $className): array
```
staabm, line 119: can be private?

```suggestion
    private function resolveSourceClassNames(string $className): array
```

staabm (changes requested):

rules/CodeQuality/Rector/Class_/AddCoversClassAttributeRector.php
maks-rafalko: done, thanks

maks-rafalko:
https://github.com/rectorphp/rector-phpunit/actions/runs/8475251813/job/23383638862?pr=319#step:5:16 - this is unrelated to the changes in this PR

Comments addressed, please re-review

staabm (approved):

samsonasik:
Rebase is needed to make CI green.

maks-rafalko:
> Rebase is needed to make CI green.

Could you please run actions? should be green now

maks-rafalko:
Green :+1: 

If you guys decide to merge it, can I ask you to make a release so we can use this new feature at Infection?

Thanks!

staabm:
@maks-rafalko maybe you could give a few notes, why this annotation is useful for projects and when one should use it?

samsonasik:
I prefer it to be configurable, and exclude from code quality set tbh, as it addition is by guessing, and may probably cause unwanted changes, as I personally not want to use it.

I will let @TomasVotruba decide

maks-rafalko:
> @maks-rafalko maybe you could give a few notes, why this annotation is useful for projects and when one should use it?

added to PR description



> and exclude from code quality set tbh, as it addition is by guessing, and may probably cause unwanted changes, as I personally not want to use it.

can we have a rule that is not part of any `set` at Rector? Cause if I remember correctly, `class-leak` was failing because initially I didn't add it anywhere.

samsonasik:
--skip-type should can be used, ref https://github.com/rectorphp/rector-src/blob/a43f37190db97fb8807cb08f4e7171a338625387/.github/workflows/code_analysis.yaml#L44

samsonasik:
double `--` should add it as argument:

```
composer class-leak -- --skipe-type="..."
```

maks-rafalko:
what about the first point?



> should I remove this new rector rule from existing set and add `--skip-type` to the place you mentioned above?

samsonasik:
Yes, imo, remove from set and skip from class-leak, it also need to be registered in rector-src MissingInSetCommand as well if it merged

maks-rafalko:
I don't get [build failure](https://github.com/rectorphp/rector-phpunit/actions/runs/8603626172/job/23575912651?pr=319#step:5:14):

```bash
vendor/bin/class-leak check config src rules \
    --skip-type="Rector\\PHPUnit\\CodeQuality\\Rector\\Class_\\AddCoversClassAttributeRector"
```

it works correctly locally. @samsonasik do you have an idea what can be wrong?

maks-rafalko:
ready for review again

maks-rafalko:
Is there anything I need to do to move it forward? @TomasVotruba @samsonasik

TomasVotruba (approved):

samsonasik:
Let's give it a try, thank you @maks-rafalko

staabm:
@maks-rafalko running the new rector on phpunit itself yields a few false positives, e.g.

```
55) tests/unit/Framework/MockObject/TestDoubleTestCase.php:21

    ---------- begin diff ----------
@@ @@
 use PHPUnit\TestFixture\MockObject\InterfaceWithReturnTypeDeclaration;
 use stdClass;

+#[\PHPUnit\Framework\Attributes\CoversClass(\PHPUnit\Framework\MockObject\TestDoubleTestCase::class)]
 abstract class TestDoubleTestCase extends TestCase
 {
     final public function testMethodReturnsNullWhenReturnValueIsNullableAndNoReturnValueIsConfigured(): void
    ----------- end diff -----------

Applied rules:
 * AddCoversClassAttributeRector


56) tests/unit/Metadata/Parser/AnnotationParserTestCase.php:41

    ---------- begin diff ----------
@@ @@
 use PHPUnit\TestFixture\Metadata\Annotation\TestDoxTest;
 use PHPUnit\TestFixture\Metadata\Annotation\UsesTest;

+#[\PHPUnit\Framework\Attributes\CoversClass(\PHPUnit\Metadata\Parser\AnnotationParserTestCase::class)]
 abstract class AnnotationParserTestCase extends TestCase
 {
     public static function provideRequiresPhpTestMethods(): array
    ----------- end diff -----------

Applied rules:
 * AddCoversClassAttributeRector


57) tests/unit/Metadata/Parser/AttributeParserTestCase.php:52

    ---------- begin diff ----------
@@ @@
 use PHPUnit\TestFixture\Metadata\Attribute\UsesTest;
 use PHPUnit\TestFixture\Metadata\Attribute\WithoutErrorHandlerTest;

+#[\PHPUnit\Framework\Attributes\CoversClass(\PHPUnit\Metadata\Parser\AttributeParserTestCase::class)]
 abstract class AttributeParserTestCase extends TestCase
 {
     #[TestDox('Parses #[BackupGlobals] attribute on class')]
    ----------- end diff -----------

Applied rules:
 * AddCoversClassAttributeRector
```

I think we should make sure that the class beeing added in the `CoversClass` annotation is not a test-class itself? wdyt?

samsonasik:
I am on mobile, it seems also needs check that the target class should also need to be in test class; eg:

```php
$isBeingTested = (bool) $this->betterNodeFinder->findFirst($node->getClassMethods(), fn(Node $subNode): bool => $subNode instanceof FullyQualified && $subNode->toString() === $targetClassName);

staabm:
my comment is addressed in https://github.com/rectorphp/rector-phpunit/pull/324

maks-rafalko:
Could you please guys give an idea when this can be released? As I understand, this package is part of `rector/rector` and we need to wait untill next `rector/rector` release?

samsonasik:
@maks-rafalko rector 1.0.5 is tagged https://github.com/rectorphp/rector/releases/tag/1.0.5

PR #320, opened 2024-03-25
Add `class-leak` to composer scripts

maks-rafalko:
- to increase its visibility
- to fix errors locally instead of seeing on GH Actions

Originally found here: https://github.com/rectorphp/rector-phpunit/pull/319#issuecomment-2017405159


samsonasik (approved):

samsonasik:
Thank you @maks-rafalko

PR #321, opened 2024-03-25
 Add `complete-check` composer script same as it's done in `rectorphp/rector-src`.

maks-rafalko:
This is a follow up for https://github.com/rectorphp/rector-phpunit/pull/320 and should be merged after it.

Add `complete-check` composer script same as it's done in `rectorphp/rector-src`.

Example taken from:

https://github.com/rectorphp/rector-src/blob/fc250dd8da281cb79f5657d2f5d6c6c0053e54d1/composer.json#L106-L111

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,6 +36,7 @@
     "scripts": {
         "phpstan": "vendor/bin/phpstan analyse --ansi --error-format symplify",
         "check-cs": "vendor/bin/ecs check --ansi",
+        "class-leak": "vendor/bin/class-leak check config src rules",
         "fix-cs": "vendor/bin/ecs check --fix --ansi",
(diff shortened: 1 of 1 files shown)


maks-rafalko:
please review @TomasVotruba @samsonasik 

It would already save me time as working on https://github.com/rectorphp/rector-phpunit/pull/319 I need o run 5 commands instead of this one :)

TomasVotruba:
Let's give it a try, thank you :)

maks-rafalko:
> Let's give it a try

but you use exactly the same approach on main rector repository :) so it's just a copy for consistency 

https://github.com/rectorphp/rector-src/blob/fc250dd8da281cb79f5657d2f5d6c6c0053e54d1/composer.json#L106-L111

PR #325, opened 2024-04-14
AddCoversClassAttributeRector: skip interfaces & traits

staabm:
`CoversClass` annotation should not be used with interfaces or traits

see 
- https://github.com/sebastianbergmann/phpunit/blob/53eebaa2a68b737b309ca2d4deb8fe961eb71200/src/Metadata/Api/CodeCoverage.php#L277-L284
- https://github.com/sebastianbergmann/phpunit/blob/53eebaa2a68b737b309ca2d4deb8fe961eb71200/src/Metadata/Api/CodeCoverage.php#L286-L293

### DIFF
--- a/rules/CodeQuality/Rector/Class_/AddCoversClassAttributeRector.php
+++ b/rules/CodeQuality/Rector/Class_/AddCoversClassAttributeRector.php
@@ -148,6 +148,11 @@ final class AddCoversClassAttributeRector extends AbstractRector
                 continue;
             }
 
+            $classReflection = $this->reflectionProvider->getClass($className);
+            if ($classReflection->isInterface()) {
+                continue;
+            }
+
             return $className;
         }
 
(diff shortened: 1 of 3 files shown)


rules/CodeQuality/Rector/Class_/AddCoversClassAttributeRector.php
```diff
@@ -148,6 +148,15 @@ private function matchExistingClassName(array $classNames): ?string
                 continue;
             }
 
+            $classReflection = $this->reflectionProvider->getClass($className);
+            if ($classReflection->isInterface()) {
+                continue;
+            }
+
+            if ($classReflection->isTrait()) {
```
samsonasik, line 156: `! $classReflection->isClass()` should be enough

staabm: there is also enum. I want to be explicit

samsonasik (approved):

samsonasik:
Thank you @staabm

PR #326, opened 2024-04-17
AddCoversClassAttributeRector: skip abstract classes

staabm:
quoting sebastian bergmann:

> This has no effect as AttributeParserTestCase is an abstract class and only metadata on concrete test classes is used.

see https://github.com/sebastianbergmann/phpunit/pull/5814#discussion_r1568275673


staabm:
//cc @maks-rafalko

samsonasik:
Thank you @staabm

PR #329, opened 2024-05-02
Fix InstanceOfComparisonRector for variables

jorgsowa:
Rule `InstanceOfComparisonRector` doesn't work when we pass a variable as the second argument to operator `instanceof`. This PR fixes it.

### DIFF
--- a/rules/CodeQuality/Rector/MethodCall/AssertInstanceOfComparisonRector.php
+++ b/rules/CodeQuality/Rector/MethodCall/AssertInstanceOfComparisonRector.php
@@ -72,9 +72,11 @@ final class AssertInstanceOfComparisonRector extends AbstractRector
         if (! $this->testsNodeAnalyzer->isPHPUnitMethodCallNames($node, $oldMethodNames)) {
             return null;
         }
+
         if ($node->isFirstClassCallable()) {
             return null;
         }
+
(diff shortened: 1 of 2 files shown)


samsonasik:
You can fix CI by run:

```
```

rules/CodeQuality/Rector/MethodCall/AssertInstanceOfComparisonRector.php
```diff
@@ -9,12 +9,15 @@
...
 use PhpParser\Node\Expr\StaticCall;
+use PhpParser\Node\Expr\Variable;
 use Rector\Exception\ShouldNotHappenException;
 use Rector\PHPUnit\NodeAnalyzer\IdentifierManipulator;
 use Rector\PHPUnit\NodeAnalyzer\TestsNodeAnalyzer;
 use Rector\Rector\AbstractRector;
 use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
 use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+use function array_keys;
+use function array_merge;
```
samsonasik, line 20: this is not part of our CS

samsonasik (approved):

samsonasik:
Thank you @jorgsowa

PR #331, opened 2024-05-27
Check if an assert method is static before calling it with self

pl-github:
In my project, I have self-defined `assert` methods that are not static because they need to access some instance variables. But the `PreferPHPUnitSelfCallRector` rule converts all method calls to static calls because the names match.

I created a patch that additionally checks if the `assert` method can be called statically.

### DIFF
--- a/src/Rector/Class_/PreferPHPUnitSelfCallRector.php
+++ b/src/Rector/Class_/PreferPHPUnitSelfCallRector.php
@@ -8,9 +8,11 @@ use PhpParser\Node;
 use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Expr\StaticCall;
 use PhpParser\Node\Stmt\Class_;
+use PHPStan\Reflection\ClassReflection;
 use PHPStan\Type\ObjectType;
 use Rector\PHPUnit\NodeAnalyzer\TestsNodeAnalyzer;
 use Rector\Rector\AbstractRector;
+use Rector\Reflection\ReflectionResolver;
 use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
 use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
 
@@ -21,6 +23,7 @@ final class PreferPHPUnitSelfCallRector extends AbstractRector
 {
     public function __construct(
         private readonly TestsNodeAnalyzer $testsNodeAnalyzer,
+        private readonly ReflectionResolver $reflectionResolver,
     ) {
     }
 
@@ -95,8 +98,17 @@ CODE_SAMPLE
                 return null;
             }
 
-            $hasChanged = true;
-            return $this->nodeFactory->createStaticCall('self', $methodName, $node->getArgs());
+            $classReflection = $this->reflectionResolver->resolveClassReflection($node);
+            if ($classReflection instanceof ClassReflection && $classReflection->hasNativeMethod($methodName)) {
+                $method = $classReflection->getNativeMethod($methodName);
+                if ($method->isStatic()) {
+                    $hasChanged = true;
+
+                    return $this->nodeFactory->createStaticCall('self', $methodName, $node->getArgs());
+                }
+            }
+
+            return null;
         });
 
         if ($hasChanged) {
--- a/tests/Rector/Class_/PreferPHPUnitSelfCallRector/Fixture/skip_non_static_assert.php.inc
+++ b/tests/Rector/Class_/PreferPHPUnitSelfCallRector/Fixture/skip_non_static_assert.php.inc
@@ -1,0 +1,18 @@
+<?php
+
+namespace Rector\PHPUnit\Tests\Rector\Class_\PreferPHPUnitSelfCallRector\Fixture;
+
+use PHPUnit\Framework\TestCase;
+
+final class SkipNonStaticAssert extends TestCase
+{
+    public function testMe()
+    {
+        $this->assertNonStatic('expected', 'actual', 'message');
+    }
+
+    public function assertNonStatic(mixed $expected, mixed $actual, string $message = '')
+    {
+        self::assertSame($expected, $actual, $message);
+    }
+}


src/Rector/Class_/PreferPHPUnitSelfCallRector.php
```diff
@@ -95,6 +98,21 @@ public function refactor(Node $node): ?Node
                 return null;
             }
 
+            $callerType = $this->nodeTypeResolver->getType($node->var);
+            if (! $callerType instanceof TypeWithClassName) {
+                return null;
+            }
+
+            $scope = $node->getAttribute(AttributeKey::SCOPE);
+            $methodReflection = $callerType->getClassReflection()?->getMethod($methodName, $scope);
```
samsonasik, line 107: better to check `hasNativeMethod()` then `getNativeMethod()` later to avoid crash on not found:

```php
$classReflection = $this->reflectionResolver->resolveClassReflectionSourceObject($node);
if ($classReflection instanceof \PHPStan\Reflection\ClassReflection && $classReflection->hasNativeMethod($methodName)) {
    $method = $classReflection->getNativeMethod($methodName);
    if (! $method->isStatic()) {
        return null;
    }
}
```

pl-github: Thank you for your suggestion. I will push an updated pull request.

src/Rector/Class_/PreferPHPUnitSelfCallRector.php
```diff
@@ -95,6 +99,19 @@ public function refactor(Node $node): ?Node
                 return null;
             }
 
+            $callerType = $this->nodeTypeResolver->getType($node->var);
+            if (! $callerType instanceof TypeWithClassName) {
+                return null;
```
samsonasik, line 104: The `$callerType` seems not needed

src/Rector/Class_/PreferPHPUnitSelfCallRector.php
```diff
@@ -95,6 +99,19 @@ public function refactor(Node $node): ?Node
...
+            }
+
+            $classReflection = $this->reflectionResolver->resolveClassReflectionSourceObject($node);
+            if ($classReflection instanceof ClassReflection && $classReflection->hasNativeMethod($methodName)) {
+                $method = $classReflection->getNativeMethod($methodName);
+                if (! $method->isStatic()) {
```
samsonasik, line 110: use positive check, then return nodefactory seems ok here, else, return null

src/Rector/Class_/PreferPHPUnitSelfCallRector.php
pl-github: Is now included in latest. version.

src/Rector/Class_/PreferPHPUnitSelfCallRector.php
pl-github: Is now included in latest. version.

src/Rector/Class_/PreferPHPUnitSelfCallRector.php
```diff
@@ -95,8 +99,21 @@ public function refactor(Node $node): ?Node
                 return null;
             }
 
-            $hasChanged = true;
-            return $this->nodeFactory->createStaticCall('self', $methodName, $node->getArgs());
+            if (! $this->nodeTypeResolver->getType($node->var) instanceof TypeWithClassName) {
+                return null;
+            }
+
+            $classReflection = $this->reflectionResolver->resolveClassReflectionSourceObject($node);
```
samsonasik, line 106: as it from `$this` variable which already validated early; the `resolveClassReflection()` 

```php

$classReflection = $this->reflectionResolver->resolveClassReflectionSourceObject($node);
```

https://github.com/rectorphp/rector-src/blob/07a61c18f678a24c102a83a4498265c874fab617/src/Reflection/ReflectionResolver.php#L66

src/Rector/Class_/PreferPHPUnitSelfCallRector.php
```diff
@@ -95,8 +98,17 @@ public function refactor(Node $node): ?Node
                 return null;
             }
 
-            $hasChanged = true;
-            return $this->nodeFactory->createStaticCall('self', $methodName, $node->getArgs());
+            $classReflection = $this->reflectionResolver->resolveClassReflectionSourceObject($node);
```
samsonasik, line 101:
```suggestion
            $classReflection = $this->reflectionResolver->resolveClassReflection($node);

Comment (samsonasik) [reply src/Rector/Class_/PreferPHPUnitSelfCallRector.php:101]: After some thinking, I think the static check only make sense if method exactly from `PHPUnit\Framework\TestCase`, that means `ClassReflection` need to be pulled from ReflectionProvider:

```php
if ($this->reflectionProvider->hasClass('PHPUnit\Framework\TestCase')) {

    $classReflection = $this->reflectionProvider->getClass('PHPUnit\Framework\TestCase');

// not has method, just skip changing...
 if (! $classReclection->hasNativeMethod($methodName)) {
    return null;
}
```

pl-github: But if I implement my own static `assert` methods in subclasses of `PHPUnit\Framework\TestCase` it would be nice if they were treated the same way as PHPUnits `assert` methods.

src/Rector/Class_/PreferPHPUnitSelfCallRector.php
pl-github: What does that mean? I do not understand.

src/Rector/Class_/PreferPHPUnitSelfCallRector.php
samsonasik: Ok :), I was thinking the original rule feature was to change to static the `assert*` method, but only from `PHPUnit\Framework\TestCase` object, 

but to keep the behavior, I am fine with it :),

Just change `resolveClassReflectionSourceObject` to `resolveClassReflection` as it validated from `this` variable early.

src/Rector/Class_/PreferPHPUnitSelfCallRector.php
samsonasik:
```suggestion
            $classReflection = $this->reflectionResolver->resolveClassReflection($node);

src/Rector/Class_/PreferPHPUnitSelfCallRector.php
pl-github: Ok, I have changed that now. 👍 

Thanks for guiding me through the Rector internals. 😄

samsonasik (approved):

Issue #344, opened 2024-06-29
NamedArgumentForDataProviderRector not working because the class DataProvider is using a scoped rector version


gnutix:
Hello there,

I've just tried the [new `NamedArgumentForDataProviderRector` rule](https://github.com/rectorphp/rector-phpunit/pull/330) and was surprised that it changed nothing in my project. I started debugging and the issue was this : 

```diff
-use RectorPrefix202406\PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\DataProvider;
```

I'm not sure if the issue is that the DataProvider class must be referenced somewhere in Rector's code to be properly handled when scoped, or if the issue is with how Rector is installed in our project (which is quite standard I think, but these things tend to get complicated real fast, so who knows...).

gnutix

TomasVotruba:
Hi, indeed, this class name should be untouched.

@samsonasik Could you fix this one, as it requires a bit of Rector internals?

samsonasik:
Yes, here the PR:

- https://github.com/rectorphp/rector-phpunit/pull/347

PR #347, opened 2024-06-30
Fix prefixed scoped DataProvider class on NamedArgumentForDataProviderRector

samsonasik:
Fixes https://github.com/rectorphp/rector-phpunit/issues/344


samsonasik:

PR #353, opened 2024-08-03
[CodeQuality] Assert compare on countable

peterfox:
# Changes

- Adds the new rule `AssertCompareOnCountableWithMethodToAssertCountRector`
- Adds a test
- Updates the docs

# Why

Based on this [PHPStan PHPUnit rule](https://github.com/phpstan/phpstan-phpunit/blob/1.4.x/src/Rules/PHPUnit/AssertRuleHelper.php) it will convert comparison on Countable objects with the count method to use the `assertCount` method instead.

Works with method and static method calls.

```diff
-$this->assertSame(1, $countable->count());
+$this->assertCount(1, $countable);
```

### DIFF
--- a/rules/CodeQuality/Rector/MethodCall/AssertCompareOnCountableWithMethodToAssertCountRector.php
+++ b/rules/CodeQuality/Rector/MethodCall/AssertCompareOnCountableWithMethodToAssertCountRector.php
@@ -9,7 +9,6 @@ use PhpParser\Node;
 use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Expr\StaticCall;
 use PhpParser\Node\Identifier;
-use PhpParser\Node\Name;
 use PHPStan\Type\ObjectType;
 use Rector\PHPUnit\NodeAnalyzer\TestsNodeAnalyzer;
 use Rector\Rector\AbstractRector;
@@ -75,7 +74,7 @@ CODE_SAMPLE
 
             if ((new ObjectType(Countable::class))->isSuperTypeOf($type)->yes()) {
                 $args = $node->getArgs();
-                $args[1] = $right->var;
+                $args[1] = new Node\Arg($right->var);
 
                 $node->args = $args;
                 $node->name = new Identifier('assertCount');
--- a/rules/CodeQuality/Rector/MethodCall/AssertCompareOnCountableWithMethodToAssertCountRector.php
+++ b/rules/CodeQuality/Rector/MethodCall/AssertCompareOnCountableWithMethodToAssertCountRector.php
@@ -77,13 +77,8 @@ CODE_SAMPLE
                 $args = $node->getArgs();
                 $args[1] = $right->var;
 
-                if ($node instanceof MethodCall) {
-                    return $this->nodeFactory->createMethodCall($node->var, 'assertCount', $args);
-                }
-
-                if ($node->class instanceof Name) {
-                    return $this->nodeFactory->createStaticCall($node->class->toString(), 'assertCount', $args);
-                }
+                $node->args = $args;
+                $node->name = new Identifier('assertCount');
             }
         }
 
--- a/rules/CodeQuality/Rector/MethodCall/AssertCompareOnCountableWithMethodToAssertCountRector.php
+++ b/rules/CodeQuality/Rector/MethodCall/AssertCompareOnCountableWithMethodToAssertCountRector.php
@@ -1,0 +1,81 @@
+<?php
+
+namespace Rector\PHPUnit\CodeQuality\Rector\MethodCall;
+
+use Countable;
+use PhpParser\Node;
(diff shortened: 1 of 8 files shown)


rules/CodeQuality/Rector/MethodCall/AssertCompareOnCountableWithMethodToAssertCountRector.php
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        if (
+            ! $node->name instanceof Identifier ||
+            ($node->name->toLowerString() !== 'assertsame' && $node->name->toLowerString() !== 'assertequals')
+        ) {
+            return null;
+        }
+
+        $right = $node->getArgs()[1]
```
samsonasik, line 64: `->isFirstClassCallable()` should be verified first and return null, then verify args total is 2

rules/CodeQuality/Rector/MethodCall/AssertCompareOnCountableWithMethodToAssertCountRector.php
```diff
@@ -0,0 +1,91 @@
...
+    public function getNodeTypes(): array
+    {
+        return [MethodCall::class, StaticCall::class];
+    }
+
+    /**
+     * @param MethodCall|StaticCall $node
+     */
+    public function refactor(Node $node): MethodCall|StaticCall|null
+    {
```
samsonasik, line 50: call is in test case class should be checked, like other rule

https://github.com/rectorphp/rector-phpunit/blob/acfcef8cfbceefee966253ef5f1b2ec57bc10e22/rules/CodeQuality/Rector/MethodCall/AssertComparisonToSpecificMethodRector.php#L94

rules/CodeQuality/Rector/MethodCall/AssertCompareOnCountableWithMethodToAssertCountRector.php
peterfox: I've added the argument count but I'm not sure what you mean by
> `->isFirstClassCallable()` should be verified first and return null

Could you clarify further?

samsonasik: `->getArgs()` got warning when there is firstclass callable inside, see

https://github.com/nikic/PHP-Parser/blob/e3f223f623980cdc074f770eb3d54aafb3af5346/lib/PhpParser/Node/Expr/CallLike.php#L32

so needs verify early:

```php
if ( ! $node->isFirstCallable()) {
    return null;
}

Comment (samsonasik) [reply rules/CodeQuality/Rector/MethodCall/AssertCompareOnCountableWithMethodToAssertCountRector.php:65]: See example

https://github.com/rectorphp/rector-phpunit/blob/acfcef8cfbceefee966253ef5f1b2ec57bc10e22/rules/CodeQuality/Rector/MethodCall/AssertNotOperatorRector.php#L66

Comment (peterfox) [reply rules/CodeQuality/Rector/MethodCall/AssertCompareOnCountableWithMethodToAssertCountRector.php:69]: @samsonasik thank you for explaining! Update made.

File: rules/CodeQuality/Rector/MethodCall/AssertCompareOnCountableWithMethodToAssertCountRector.php [hunk 3]
```diff
@@ -0,0 +1,91 @@
...
+
+        if (
+            ($right instanceof MethodCall)
+            && $right->name instanceof Identifier
+            && $right->name->toLowerString() === 'count'
+            && $right->getArgs() === []
+        ) {
+            $type = $this->getType($right->var);
+
+            if ((new ObjectType(Countable::class))->isSuperTypeOf($type)->yes()) {
```
samsonasik, line 80: use normal `"Countable"` string, even usually native php class not prefixed by php scoper, just to avoid issue on downgrade:

```suggestion
            if ((new ObjectType('Countable'))->isSuperTypeOf($type)->yes()) {
```

rules/CodeQuality/Rector/MethodCall/AssertCompareOnCountableWithMethodToAssertCountRector.php
```diff
@@ -0,0 +1,91 @@
...
+class AssertCompareOnCountableWithMethodToAssertCountRector extends AbstractRector
+{
+    public function __construct(
+        private readonly TestsNodeAnalyzer $testsNodeAnalyzer
+    ) {
+    }
+
+    public function getRuleDefinition(): RuleDefinition
+    {
+        return new RuleDefinition('', [
```
samsonasik, line 31: rule definition string is needed

samsonasik (approved):

TomasVotruba (approved):

TomasVotruba:
Thank you 😊

PR #360, opened 2024-09-22
Move WithConsecutiveRector to its PHPUnit100 namespace, CreateMockToAnonymousClassRector + PreferPHPUnitSelfCallRector to CodeQuality

TomasVotruba:
Just noticed this when the rule could not be found:

https://getrector.com/find-rule?query=withConsecutive

It was in wrong location, should be always in `/rules`


TomasVotruba:
Risk of BC break is present, but these are part of sets, or used as onesies - run once and then removed. So we should be fine :)

smichaelsen:
Ran into this as a breaking change. But was only a 5 minutes inconvenience until I figured it out. Thanks for your work!

PR #373, opened 2024-09-29
Ensure check isFirstClassCallable() before getArgs() call on CallLike

samsonasik:
to avoid assert error on first class callable.


samsonasik:

PR #388, opened 2024-11-03
[code-quality] Add NarrowUnusedSetUpDefinedPropertyRector

TomasVotruba:
- **misc**
- **[code-quality] Add NarrowUnusedSetUpDefinedPropertyRector**

### DIFF
--- a/rules/CodeQuality/Rector/Class_/NarrowUnusedSetUpDefinedPropertyRector.php
+++ b/rules/CodeQuality/Rector/Class_/NarrowUnusedSetUpDefinedPropertyRector.php
@@ -1,0 +1,183 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\PHPUnit\CodeQuality\Rector\Class_;
+
+use PhpParser\Node;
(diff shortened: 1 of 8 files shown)


rules/CodeQuality/Rector/Class_/NarrowUnusedSetUpDefinedPropertyRector.php
```diff
@@ -0,0 +1,183 @@
...
+                continue;
+            }
+
+            if (! $this->isName($property->type, self::MOCK_OBJECT_CLASS)) {
+                continue;
+            }
+
+            if (! $this->isPropertyUsedOutsideSetUpClassMethod($node, $setUpClassMethod, $property)) {
+                continue;
+            }
```
samsonasik, line 113: the check is flipped, continue should check when property is used

samsonasik: ok, it seems fixed at https://github.com/rectorphp/rector-phpunit/commit/f24bef525133df61ef556e34965f45625c8215c6

PR #390, opened 2024-11-03
[CodeQuality] Fix flipped check condition only remove property when unused on NarrowUnusedSetUpDefinedPropertyRector

samsonasik:
see https://github.com/rectorphp/rector-phpunit/pull/388#pullrequestreview-2411971619


samsonasik:
it seems covered at https://github.com/rectorphp/rector-phpunit/commit/f24bef525133df61ef556e34965f45625c8215c6

PR #391, opened 2024-11-03
[CodeQuality] Skip non private on non final class on NarrowUnusedSetUpDefinedPropertyRector

samsonasik:
when class is not final, and property is non-private, it can be used by child class.


samsonasik:

PR #392, opened 2024-11-04
[AnnotationsToAttributes] Fix CoversMethod usage on CoversAnnotationWithValueToAttributeRector

samsonasik:
CoversMethod exists by direct commit at https://github.com/sebastianbergmann/phpunit/commit/fa8db04e2daefdf77075e3388d2512aea549470a

Also fix tests which should test covers method, not class.


samsonasik:
oh, it seems removed after it, it no longer exists in latest phpunit version

PR #399, opened 2024-11-04
[CodeQuality] Skip direct use without assign on NarrowUnusedSetUpDefinedPropertyRector

samsonasik:
It currently cause undefined variable error, ref https://getrector.com/demo/b2a25b32-fae8-45e1-87d8-2a8946ff701d


samsonasik:
it seems already covered at https://github.com/rectorphp/rector-phpunit/pull/398

PR #403, opened 2024-11-08
[PHPUnit 10] Skip with Magic __get() on AssertIssetToAssertObjectHasPropertyRector

samsonasik:
with magic `__get()` property doesn't need to exists to be called.

see https://3v4l.org/3Pckj


samsonasik:

PR #405, opened 2024-11-10
[PHPUnit 10] Remove AssertIssetToAssertObjectHasPropertyRector

samsonasik:
This rule is apply invalid transformation, since `isset()` is verify if property has null value as well, see https://3v4l.org/KBkr3

so the transformation is invalid.

For transform to `assertObjectHasProperty`, there is already configured config for that, see

https://github.com/rectorphp/rector-phpunit/blob/687747d8c7f161995dd2327b5c721a4090929e69/config/sets/phpunit100.php#L32-L34

which transformable from `AssertPropertyExistsRector` rule for that, see

https://github.com/rectorphp/rector-phpunit/blob/687747d8c7f161995dd2327b5c721a4090929e69/rules/CodeQuality/Rector/MethodCall/AssertPropertyExistsRector.php#L23

https://github.com/rectorphp/rector-phpunit/blob/687747d8c7f161995dd2327b5c721a4090929e69/rules-tests/CodeQuality/Rector/MethodCall/AssertPropertyExistsRector/Fixture/fixture2.php.inc#L5-L27

so combine them will apply the syntax transformation.


samsonasik:
@TomasVotruba Ready to merge 👍

TomasVotruba:
Makes sense, thank you ☺️

PR #414, opened 2024-12-09
YieldDataProviderRector: added failling test

staabm:
I think it would be useful, if rector transforms a `array` return type into a `Iterator`, that it also fixes a possible `@return` phpdoc type of a dataprovider (otherwise PHPStan is starting to yell about incompatible native vs. phpdoc return-type

### DIFF
--- a/rules-tests/CodeQuality/Rector/Class_/YieldDataProviderRector/Fixture/use_data_provider_with_phpdoc.php.inc
+++ b/rules-tests/CodeQuality/Rector/Class_/YieldDataProviderRector/Fixture/use_data_provider_with_phpdoc.php.inc
@@ -37,7 +37,7 @@ namespace Rector\PHPUnit\Tests\CodeQuality\Rector\Class_\YieldDataProviderRector
 
 use PHPUnit\Framework\TestCase;
 
-final class UseDataProviderTest extends TestCase
+final class UseDataProviderTestPhpdoc extends TestCase
 {
     #[\PHPUnit\Framework\Attributes\DataProvider('dataProvider')]
     #[\PHPUnit\Framework\Attributes\DataProvider('provideDataForProvider')]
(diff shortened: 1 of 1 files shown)


rules-tests/CodeQuality/Rector/Class_/YieldDataProviderRector/Fixture/use_data_provider_with_phpdoc.php.inc
```diff
@@ -0,0 +1,61 @@
+<?php
+
+namespace Rector\PHPUnit\Tests\CodeQuality\Rector\Class_\YieldDataProviderRector\Fixture;
+
+use PHPUnit\Framework\TestCase;
+
+final class UseDataProviderTest extends TestCase
```
samsonasik, line 7:
```suggestion
final class UseDataProviderTestPhpdoc extends TestCase
```

staabm:
@samsonasik could you have a look into the rule?

rules-tests/CodeQuality/Rector/Class_/YieldDataProviderRector/Fixture/use_data_provider_with_phpdoc.php.inc
```diff
@@ -0,0 +1,61 @@
...
+
+    /** @return array<int, array<string>> */
+    public static function provideDataForProvider()
+    {
+        return [
+            ['<?php implode("", $foo, );', '<?php implode($foo, "", );']
+        ];
+    }
+
+    public static function dataProvider()
```
staabm, line 23: At this provider we could add another case with `list<...>`

samsonasik:
@staabm I cherry-picked your commit at PR 

- https://github.com/rectorphp/rector-phpunit/pull/415

PR #415, opened 2024-12-09
[PHPUnit] Change @return array<> to @return Iterator<> on YieldDataProviderRector

samsonasik:
Closes https://github.com/rectorphp/rector-phpunit/pull/414


samsonasik:

PR #424, opened 2024-12-23
Improve AssertCompareOnCountableWithMethodToAssertCountRector

VincentLanglet:
Previously
- It supported only assertEquals/assertSame
- It supported only count function

Now
- It also supports assertNotEquals/assertNotSame
- It also supports sizeof and iterator_count functions


TomasVotruba (approved):

TomasVotruba:
Thank you for the improvements :partying_face:

PR #425, opened 2024-12-23
Introduce AssertClassNameComparisonToInstanceOfRector

VincentLanglet:


TomasVotruba:
Just checking the existing rules, it might be worth including in `AssertInstanceOfComparisonRector`. What do you think?

VincentLanglet:
> Just checking the existing rules, it might be worth including in `AssertInstanceOfComparisonRector`. What do you think?
> 
> To avoid too granulate rules doing almost the same thing

Sure

PR #427, opened 2024-12-23
Remove AssertCompareToSpecificMethodRector from CodeQuality set

VincentLanglet:


TomasVotruba:
Thanks :+1:  I'm looking at the rule now and see there is on reason for its existance. I'll handle the deprecatoin

PR #430, opened 2024-12-25
[CodeQuality] Skip loose compare float vs integer on AssertEqualsToSameRector

samsonasik:
The following should be skipped:

```php
$this->assertEquals(1.0, 1);
```

as cause error on assertSame().


samsonasik:
on `assertSame()`, I think it should be skipped if second arg is mixed, as it never guessable.

samsonasik:
Ok, comparing array or bool to non-bool should not happen on test, but float/integer to integer/float may happen.

This patch only skip:

- compare float to integer
- composer integer to float
- composer float to mixed
- composer integer to mixed

samsonasik:
All checks have passed 🎉 @TomasVotruba I am merging it to have faster feedback to test ;)

PR #436, opened 2024-12-26
[CodeQuality] Skip loose comparison on assertNotEquals as well on AssertEqualsToSameRector

samsonasik:


rules-tests/CodeQuality/Rector/MethodCall/AssertEqualsToSameRector/Fixture/skip_loose_compare_assert_not_equals.php.inc
```diff
@@ -0,0 +1,13 @@
...
+
+use PHPUnit\Framework\TestCase;
+
+final class SkipLooseCompareAssertNotEquals extends TestCase
+{
+    public function test(float $float)
+    {
+        $this->assertNotEquals(1, $float);
+    }
+}
```
TomasVotruba, line 13: This should be covered a separate rule that would update both name and type.

```diff
-$this->assertNotEquals(1, $float);
+$this->assertNotSame(1.0, $float);
```

samsonasik: loose comparison needs to be verified manually, eg int vs float, it can be true while on assertSame it be false, on opposite, it happen as well, make always not same, which actually can be equal

samsonasik: Oh, I see what you mean, that's value need to be changed, however, that's need to be checked as constant value, otherwise, cast `(float)/(int)` may be needed

TomasVotruba: Yes, it should be applied on constant int/float values only. I always handle this change manually, often ~100s of cases and it's such a pain :D

PR #440, opened 2025-01-15
Added testcase to demonstrate rector#8977

andrewnicols:
Where an inline docblock is used in another tag, all subsequent tags are removed.

https://github.com/rectorphp/rector/issues/8977

I've tried to keep this as small as possible but have deliberately incldued:
* a text node
* a tag before the tag containing the inline tag
* a tag after the tag containing the inline tag
* the tag we want to convert to attribute

### DIFF
--- a/tests/AnnotationsToAttributes/Fixture/inline_docblock_tag.php.inc
+++ b/tests/AnnotationsToAttributes/Fixture/inline_docblock_tag.php.inc
@@ -1,0 +1,27 @@
+<?php
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @copyright Example {@link https://example.com}
+ * @covers \Tests\BarController
+ */
+class BarController extends TestCase
+{
+}
+
+?>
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Thanks! Are all the docblock lines needed to trigger the bug? I've just tried to remove the first one, and bug remains.
The less code there is the better, as in docblock parser every line and every element is split into token we have to debug.

samsonasik:
@andrewnicols please remove these commited files:

- `rules-tests/CodeQuality/Rector/Class_/YieldDataProviderRector/Fixture/should_not_remove_doc_blocks.php`
- `rules-tests/CodeQuality/Rector/MethodCall/UseSpecificWillMethodRector/Fixture/will.php`

andrewnicols:
Done - Not sure how those got there. I literally cloned the repo to create these tests.

samsonasik:
Thank you @andrewnicols

PR #447, opened 2025-01-21
Add str_contains() support to AssertTrueFalseToSpecificMethodRector

TomasVotruba:


rules/CodeQuality/Rector/MethodCall/AssertTrueFalseToSpecificMethodRector.php
```diff
@@ -214,7 +215,8 @@ private function buildNewArguments(
             return [...$funcCallOrEmptyNodeArgs, ...$oldArguments];
         }
 
-        if ($funcCallOrEmptyNodeName === 'is_a') {
+        if (in_array($funcCallOrEmptyNodeName, ['is_a', 'str_contains'])) {
```
samsonasik, line 218: add true to 3rd argument:

```suggestion
        if (in_array($funcCallOrEmptyNodeName, ['is_a', 'str_contains'], true)) {
```

PR #451, opened 2025-01-29
[AnnotationsToAttributes] Fix @covers qualified::method conversion

andrewnicols:
The PHPUnit 9.6 `@covers` annotation supports specification of methods:
https://docs.phpunit.de/en/9.6/annotations.html#covers

That is:

```
/**
 * @covers \someClass::someMethod
 */
```

The `CoversAnnotationWithValueToAttributeRector` conversion is currently unaware of this and converting it to a `CoversClass` attribute, that is:

```
#[\PHPUnit\Framework\Attributes\CoversClass(\someClass::someMethod)
```

This is, of course, a syntax error. It should create a `CoversMethod` attribute:

```
#[\PHPUnit\Framework\Attributes\CoversClass(\someClass::class, 'someMethod')
```

This patch changes the behaviour to look for a `::` in the annotation and, if found, grab the class and use the `CoversMethod` Attribute with the class and method name. The behaviour for an annotation that _starts_ with `::` is unaffected.

### DIFF
--- a/rules-tests/AnnotationsToAttributes/Rector/Class_/CoversAnnotationWithValueToAttributeRector/Fixture/covers_class.php.inc
+++ b/rules-tests/AnnotationsToAttributes/Rector/Class_/CoversAnnotationWithValueToAttributeRector/Fixture/covers_class.php.inc
@@ -6,6 +6,7 @@ use PHPUnit\Framework\TestCase;
 
 /**
  * @covers \Rector\PHPUnit\Tests\AnnotationsToAttributes\Rector\Class_\CoversAnnotationWithValueToAttributeRector\Source\ExistingClass
+ * @covers \Rector\PHPUnit\Tests\AnnotationsToAttributes\Rector\Class_\CoversAnnotationWithValueToAttributeRector\Source\AnotherExistingClass::someMethod
  */
 final class CoversClass extends TestCase
 {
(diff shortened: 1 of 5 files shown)


samsonasik:
Please run:

```
```

to fix CI

andrewnicols:
Apologies @samsonasik, I thought I'd created this as a draft request. I've cleaned up coding style now.

samsonasik:
did you forget git push?

andrewnicols:
Indeed I did. It's one of 'those' days.

samsonasik (approved):

samsonasik:
I don't follow latest Covers* in PHPUnit for class method, last time old days, it was not exists, or I was looking to different things.

If you already verify it, let's give it a try :)

samsonasik:
Thank you @andrewnicols

andrewnicols:
It looks like the `CoversMethod` attribute was added in PHPUnit 11. It was not present in PHPUnit 10, but there is no alternative in that version of PHPUnit.

- PHPUnit 11: https://docs.phpunit.de/en/11.5/attributes.html#coversmethod
- PHPunit 10: https://docs.phpunit.de/en/10.5/attributes.html#coversclass
- PHPUnit 9.6 annotation: https://docs.phpunit.de/en/9.6/annotations.html#covers

Given there is no ability to specify this in PHPUnit 10, and removing the method altogether would cause data loss that is the loss of the coverage information) there are some options:
- Use `CoversMethod` on it's own - this will work for PHPUnit 11, but for PHPUnit 10 the coverage data will be 'lost'
- Use `CoversClass` on its own always - this will work for PHPUnit 10+, but may give more specific coverage data than desired
- Use both `CoversMethod` and `CoversClass` - this is a bit pointless because the Method coveage is already covered by the class.

samsonasik:
Ugh, what a mess, it seems ok with coverage loss, as unknown class attribute seems not error, see https://3v4l.org/iqNL1 except you have other option

PR #456, opened 2025-02-08
Adds AssertCountWithZeroToAssertEmptyRector rule

peterfox:
# Changes

* Adds a new AssertCountWithZeroToAssertEmptyRector.
* Adds tests for the rule.
* Added to code quality set.

# Why

This is just a small but useful rector rule. It will convert use of `assertCount` with a first parameter of zero to be `assertEmpty`. Also handles the `not` use as well.

```diff
-$this->assertCount(0, $array);
+$this->assertEmpty($array);
-$this->assertNotCount(0, $array);
+$this->assertNotEmpty($array);
```

### DIFF
--- a/config/sets/phpunit-code-quality.php
+++ b/config/sets/phpunit-code-quality.php
@@ -15,6 +15,7 @@ use Rector\PHPUnit\CodeQuality\Rector\ClassMethod\ReplaceTestAnnotationWithPrefi
 use Rector\PHPUnit\CodeQuality\Rector\Foreach_\SimplifyForeachInstanceOfRector;
 use Rector\PHPUnit\CodeQuality\Rector\MethodCall\AssertCompareOnCountableWithMethodToAssertCountRector;
 use Rector\PHPUnit\CodeQuality\Rector\MethodCall\AssertComparisonToSpecificMethodRector;
+use Rector\PHPUnit\CodeQuality\Rector\MethodCall\AssertCountWithZeroToAssertEmptyRector;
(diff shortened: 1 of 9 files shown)


TomasVotruba:
Thanks Peter, this looks good :+1: 

We could add it to the code quality set, what do you think?

peterfox:
Will do @TomasVotruba !

TomasVotruba:
Lets ship it 😉 

Thank yoi

samsonasik:
@peterfox It seems buggy on combine with `AssertEmptyNullableObjectToAssertInstanceofRector`

```diff
-        $this->assertCount(0, $this->bar->baz());
+        $this->assertNotInstanceOf(Collection::class, $this->bar->baz());
     }
    ----------- end diff -----------

Applied rules:
 * AssertCountWithZeroToAssertEmptyRector
 * AssertEmptyNullableObjectToAssertInstanceofRector
```

that cause error:

```
Failed asserting that an object is not an instance of interface Doctrine\Common\Collections\Collection.
```

I will check if that resolvable

samsonasik:
@peterfox I created new PR to skip union type on : 

- https://github.com/rectorphp/rector-phpunit/pull/459

TomasVotruba:
I'm looking into this rule once again, as it actually creates less readable code. We ignore it compleetely on couple of projects.
`empty()` is too vague structure and PHPStan and Rector rules try to refactor away form it.

Also in `$this->assertCount(x, ...)` is more readable in context of similar counter tests:

```php
$this->assertCount(1, ...);
$this->assertCount(5, ...);
$this->assertCount(0, ...);
```

Saying that, I'll deprecate and remove this rule form the set, to promote higher quality code.

Thanks for understanding :+1:

PR #466, opened 2025-03-11
[CodeQuality] Add AddInstanceofAssertForNullableInstanceRector

TomasVotruba:


samsonasik:
It seems buggy that cause undefined variable 

```
Undefined variable $legIdentifier
```

I will look into it.

samsonasik:
https://github.com/rectorphp/rector-phpunit/pull/470

PR #469, opened 2025-03-23
[code-quality] Add RemoveDataProviderParamKeysRector

TomasVotruba:


samsonasik:
It seems buggy on yield when code quality set used in phpunit 10, it probably related with how version usage, or phpunit version check is needed on this rule:

```
PHP Warning:  Undefined array key "fuelOrders"
```

I will look into it.

TomasVotruba:
The code quality set promotes best practices and opinionated saint defaults. Often these keys are added randomly to only couple data providers.

Also, the param name - string key is unnecessary complexity. The order is good enough.

If someone is using these keys consistently and on purpose, this rule can be skipped.

PR #470, opened 2025-03-29
[CodeQuality] Handle crash on multiple assert cause undefined variable on AddInstanceofAssertForNullableInstanceRector

samsonasik:
Ref https://github.com/rectorphp/rector-phpunit/pull/466#issuecomment-2762717039
Ref https://getrector.com/demo/b6a49070-ed48-44a0-885c-54dabbfdbf22

second assert is inserted before variable definition so cause undefined variable

```
Undefined variable $someObject2
```


samsonasik:
Fixed :tada:

PR #471, opened 2025-03-31
[CodeQuality] fix assertInstanceOf casing

staabm:
> Call to method PHPUnit\Framework\Assert::assertInstanceOf() with incorrect case: assertInstanceof

grafik


samsonasik:
Thank you @staabm

PR #472, opened 2025-03-31
Fix RemoveNamedArgsInDataProviderRector so that it is not applied if no change is made

carlos-granados:
Also changes the name of the fixture file to name it `skip_...` and changes it so that the changed part is removed as no change is expected

I also changed the name of the `handleStmt` function as that function was not handling a statement, instead it was handling an `Array_` node


rules-tests/PHPUnit100/Rector/Class_/RemoveNamedArgsInDataProviderRector/Fixture/skip_provide_data_with_no_named_args.php.inc
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace Rector\PHPUnit\Tests\PHPUnit100\Rector\Class_\RemoveNamedArgsInDataProviderRector\Fixture;
+
+use PHPUnit\Framework\TestCase;
+
+final class ProvideDataWithNoNamedArgs extends TestCase
```
samsonasik, line 7:
```suggestion
final class SkipProvideDataWithNoNamedArgs extends TestCase
```

carlos-granados: Updated

samsonasik:
Thank you @carlos-granados

PR #489, opened 2025-05-17
[CodeQuality] Skip with if else condition result on AddInstanceofAssertForNullableInstanceRector

samsonasik:


samsonasik:

PR #490, opened 2025-05-20
[PHPUnit 10] Drop ParentTestClassConstructorRector

samsonasik:
constructor is marked as final since phpunit 12.0.3 so I think the rule `ParentTestClassConstructorRector` itself can be removed.

Ref phpunit diff https://github.com/sebastianbergmann/phpunit/commit/3263f4c62d4af44777ff1c81d093ee88eafccdad

Closes https://github.com/rectorphp/rector/issues/9180


samsonasik:
@Jean85 this should fix it :)

samsonasik:
@TomasVotruba ready to merge 👍

TomasVotruba:
This rule is for PHPUnit 10 only. It must stay to help with upgrade of PHPUnit 9 to 10. It should not be run on newer PHPUnit

samsonasik:
so we need another rule for phpunit 12 to drop `__construct` since it marked as final on phpunit 12?

If yes, I think "other rule" on the list check to be ignored is needed to avoid flip flop add `__construct` if both rules executed.

samsonasik:
Ok, I think flag like `$node->setAttribute('hasRemovedFinalConstruct', true);` can be added on phpunit 12 rule so on phpunit 10 rule can be skipped if has `hasRemovedFinalConstruct` flag.

samsonasik:
I created new PR for it:

-  https://github.com/rectorphp/rector-phpunit/pull/491

PR #491, opened 2025-05-20
 [PHPUnit 12] Add RemoveOverrideFinalConstructTestCaseRector 

samsonasik:
- add phpunit 12 rule to remove `__construct` as already final
- use attribute flag `hasRemovedFinalConstruct` so in rule `ParentTestClassConstructorRector` can be checked to avoid flip flop change when both rules accidentally registered.
- Register missing phpunit 11 rule on config.

Fixes Closes https://github.com/rectorphp/rector/issues/9180

### DIFF
--- a/config/sets/phpunit110.php
+++ b/config/sets/phpunit110.php
@@ -3,6 +3,8 @@
 declare(strict_types=1);
 
 use Rector\Config\RectorConfig;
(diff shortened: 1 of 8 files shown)


samsonasik:
Fixed 🎉 /cc @Jean85

samsonasik:

TomasVotruba (approved):

TomasVotruba:
Looks good, thank you 👍

PR #492, opened 2025-05-20
[AnnotationsToAttributes] Apply to class level on @uses on AnnotationWithValueToAttributeRector

samsonasik:
Make rule configurable to apply on class level.

Fixes https://github.com/rectorphp/rector/issues/9178

### DIFF
--- a/rules/AnnotationsToAttributes/Rector/Class_/AnnotationWithValueToAttributeRector.php
+++ b/rules/AnnotationsToAttributes/Rector/Class_/AnnotationWithValueToAttributeRector.php
@@ -133,7 +133,7 @@ CODE_SAMPLE
                     [$attributeValue]
                 );
 
-                if ($node instanceof ClassMethod && $annotationWithValueToAttribute->getIsOnClassLevel()) {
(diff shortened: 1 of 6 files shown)


samsonasik:
Fixed 🎉 /cc @staabm

samsonasik:

staabm (commented):
thanks for the fix. I think we need the same behaviour on `UsesFunction` annotation which also is only supported on class-level

https://github.com/sebastianbergmann/phpunit/blob/6dbcca2e8c33ffe66182ad586572f22702b47967/src/Framework/Attributes/UsesFunction.php#L19

rules/AnnotationsToAttributes/Rector/Class_/AnnotationWithValueToAttributeRector.php
```diff
@@ -101,6 +103,10 @@ public function refactor(Node $node): ?Node
             return null;
         }
 
+        if ($node instanceof Class_) {
```
staabm, line 106: do we need to skip anonymous classes here?

TomasVotruba (approved):

PR #493, opened 2025-05-22
PHPUnit 10/11 Only set attribute when target attribute class exists on CoversAnnotationWithValueToAttributeRector

samsonasik:
Fixes https://github.com/rectorphp/rector/issues/9183


samsonasik:
@TimelessTron this should fix https://github.com/rectorphp/rector/issues/9183

samsonasik:
@TomasVotruba let's merge to have faster feedback to test ;)

PR #494, opened 2025-05-22
Only remove @covers when target exists

samsonasik:
@TimelessTron this should ensure only remove when attribute group added.

Ref https://github.com/rectorphp/rector/issues/9183#issuecomment-2900116246

### DIFF
--- a/rules/AnnotationsToAttributes/Rector/Class_/CoversAnnotationWithValueToAttributeRector.php
+++ b/rules/AnnotationsToAttributes/Rector/Class_/CoversAnnotationWithValueToAttributeRector.php
@@ -303,7 +303,9 @@ CODE_SAMPLE
                 continue;
             }
 
-            if (str_contains($desiredTagValueNode->value->value, '::') && ! $this->reflectionProvider->hasClass(self::COVERS_METHOD_ATTRIBUTE)) {
+            if (str_contains($desiredTagValueNode->value->value, '::') && ! $this->reflectionProvider->hasClass(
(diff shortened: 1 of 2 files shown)


samsonasik:
It seems cause phpunit notice:

https://github.com/rectorphp/rector-phpunit/actions/runs/15180236245/job/42688052509?pr=494#step:5:35

@TimelessTron could you verify the use of `@covers ::__construct` on class level is correct, if no, we can just remove it as it won't be changed to attribute anyway

samsonasik:
Let's remove `@covers ::__construct` on class level, it should be on method if needed.

rules/AnnotationsToAttributes/Rector/Class_/CoversAnnotationWithValueToAttributeRector.php
```diff
@@ -304,6 +303,10 @@ private function removeMethodCoversAnnotations(ClassMethod $classMethod): bool
                 continue;
             }
 
+            if (str_contains($desiredTagValueNode->value->value, '::') && ! $this->reflectionProvider->hasClass(self::COVERS_METHOD_ATTRIBUTE)) {
+                continue;
+            }
```
samsonasik: this one seems can cover it, I am thinking of moving `CoversTrait` and `CoversMethod` to phpunit 11 rule tho, but that will be need in separate PR

samsonasik:
Let's merge so we have faster feedback.

PR #496, opened 2025-05-29
[CodeQuality] Skip in static arrow function on PreferPHPUnitThisCallRector

samsonasik:
Fixes https://github.com/rectorphp/rector/issues/9199


samsonasik:
Fixed 🎉 /cc @philiprabbett

PR #498, opened 2025-06-25
[code-quality] Add StringCastAssertStringContainsStringRector

TomasVotruba:


rules-tests/CodeQuality/Rector/MethodCall/AssertTrueFalseToSpecificMethodRector/Fixture/fixture2.php
```diff
@@ -0,0 +1,16 @@
+<?php
```
mttsch, line 1: @TomasVotruba Was this file added intentionally? Due to its missing `.inc` extension, it is always removed after running tests locally.

samsonasik: That seems due to run phpunit, and cut in the middle of process, then next is committed

Issue #499, opened 2025-07-02
`DataProviderAnnotationToAttributeRector` breaks PHPDoc


cs278:
Applying `DataProviderAnnotationToAttributeRector` can break PHPDoc blocks:

```php
<?php

declare(strict_types=1);

use PHPUnit\Framework\TestCase;

final class RectorTest extends TestCase
{
    /**
     * @dataProvider dataProvider
     *
     * This is a useless test to demonstrate a problem.
     */
    public function testProvider(bool $expected): void
    {
        self::assertSame($expected, $expected);
    }

    public static function dataProvider(): iterable
    {
        yield [true];
    }
}
```

```php
<?php

declare(strict_types=1);

use PHPUnit\Framework\TestCase;

final class RectorTest extends TestCase
{
    #[\PHPUnit\Framework\Attributes\DataProvider('dataProvider
This is a useless test to demonstrate a problem.')]
    public function testProvider(bool $expected): void
    {
        self::assertSame($expected, $expected);
    }

    public static function dataProvider(): iterable
    {
        yield [true];
    }
}
```

I think you could argue that the docblock is actually malformed as the annotation is supposed come after the description but given there isn't a definitive (to my knowledge) specification it's probably best not to assume that.

TomasVotruba:
Thank you for your report! Seems valid bug.

We'll need an isolated failing demo link from: http://getrector.com/demo,
that way we can reproduce the bug.

cs278:
No problem! https://getrector.com/demo/e2b8715f-0195-4fa6-8b96-06e11602ac0c

TomasVotruba:
@cs278 Thanks 👍 

@samsonasik Could you look into this one? The comment should not become part of the attribute.

samsonasik:
comment below phpdoc is technically invalid, it should be placed before, so instead of:

```php
    /**
     * @dataProvider dataProvider
     *
     * This is a useless test to demonstrate a problem.
     */
```

as that became part of the phpdoc, **use this below instead**:

↓

```php
    /**
     * This is a useless test to demonstrate a problem.
     *
     * @dataProvider dataProvider
     */
```

the tweak may can be used for it, but that's will be buggy tho

TomasVotruba:
Still, I'd expect the comment to remain comment, without any changes. 
How does phpdoc-parser deal with this?

samsonasik:
@cs278 @TomasVotruba I created PR : 

[...]

PR #500, opened 2025-07-02
[Attribute] Handle comment below @dataProvider doc on DataProviderAnnotationToAttributeRector

samsonasik:
Fixes https://github.com/rectorphp/rector-phpunit/issues/499

when comment is below `@dataProvider`, it can be kept, when below, it marked as part of it, so it removed.


TomasVotruba:
Great work 👌

PR #505, opened 2025-08-07
Cover array shape iterator for YieldDataProviderRector+RemoveUselessReturnTagRector.

biozshock:
Test to identify the issue.

As far as i was able to debug the issue is actually in the doc parsing in the phpstan.
In this case the `$variances` [here](https://github.com/phpstan/phpstan-src/blame/2.1.22/src/PhpDoc/TypeNodeResolver.php#L873) is an empty array.
Though the PHPStan itself processes the "before" and "after" correctly.


biozshock:
There's a rector change in the other part of the code. Should it be fixed here, or in a separate PR?

samsonasik:
Run here seems fine as only single file

samsonasik:
The issue seems on rector-src on `ObjectTypeMapper`

https://github.com/rectorphp/rector-src/blob/ac4f09b4f882188d21b3fa13f6b654a2373b3545/src/PHPStanStaticTypeMapper/TypeMapper/ObjectTypeMapper.php#L54-L56

samsonasik:
I pushed empty commit to trigger latest rector-src https://github.com/rectorphp/rector-phpunit/pull/505/commits/55220b80326d39eb792c1107dc01707f8df93f3a 👍

PR #510, opened 2025-08-19
[code-quality] Add MatchAssertSameExpectedTypeRector

TomasVotruba:

PR #513, opened 2025-08-19
Use Int_ over deprecated LNumber

samsonasik:
LNumber is deprecated, use `Int_` instead, ref https://github.com/rectorphp/rector-phpunit/pull/510#pullrequestreview-3131899805


samsonasik:

TomasVotruba:
I wonder why PHPStan didn't pick it up

TomasVotruba:
That's another magic not supported. PHPStan can pick up only annotation right above the class.

samsonasik:
I may can see if I have a chance to create PR to nikic/php-parser to add `@deprecated` above the `LNumber` class and other flagged as `@deprecated` so PHPStan can pick it up.

samsonasik:
https://github.com/nikic/PHP-Parser/pull/1103

PR #519, opened 2025-08-21
Revert rule removal of MatchAssertSameExpectedTypeRector

samsonasik:
I will add test for invalid test after combined with `AssertEqualsToSameRector`


samsonasik:
Test added https://github.com/rectorphp/rector-phpunit/pull/519/commits/a59c74644fa61ce6babbef02277abd92a076e4ae

Here the

PR #520, opened 2025-08-22
 [Alternative 1] Don't change to assertSame() on numeric string on AssertEqualsToSameRector

samsonasik:
@TomasVotruba this is alternative solution to not change to `assertSame` on numeric string for failure test https://github.com/rectorphp/rector-phpunit/actions/runs/17128495249/job/48586074883#step:5:43


TomasVotruba:
Looks good :+1:  I recall we use this approach in some other rules too, good job

PR #522, opened 2025-08-22
[CodeQuality] Skip assertEquals() on integer to numeric string on MatchAssertSameExpectedTypeRector

samsonasik:
@TomasVotruba per https://github.com/rectorphp/rector-phpunit/pull/520#issuecomment-3214192620 I tested in our project, the change on `assertEquals` still invalid on `"123"` vs `"0000000000000000000123"`

```diff
--- Expected
+++ Actual
@@ @@
-'123'
+'0000000000000000000123'
```

This patch skip `assertEquals` on convert. 

The converting `string` to `int` also needs consideration tho, but that when I found another error.


samsonasik:
All checks have passed 🎉 @TomasVotruba I am merging it to have faster feedback to test in our project ;)

TomasVotruba:
I see, it's because of internal casting of PHP. Thanks for fix :+1:

PR #528, opened 2025-09-01
[code-quality] Add AddReturnTypeToDependedRector

TomasVotruba:


rules/CodeQuality/Rector/Class_/AddReturnTypeToDependedRector.php
```diff
@@ -0,0 +1,149 @@
...
+use PhpParser\Node;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Stmt\Class_;
+use Rector\PhpParser\Node\BetterNodeFinder;
+use Rector\PHPStanStaticTypeMapper\Enum\TypeKind;
+use Rector\PHPUnit\NodeAnalyzer\TestsNodeAnalyzer;
+use Rector\Rector\AbstractRector;
+use Rector\StaticTypeMapper\StaticTypeMapper;
+use Rector\TypeDeclaration\TypeInferer\SilentVoidResolver;
+use Rector\TypePerfect\NodeFinder\ReturnNodeFinder;
```
samsonasik, line 16: `Rector\TypePerfect` is not part of rector/rector

samsonasik (changes requested):

rules/CodeQuality/Rector/Class_/AddReturnTypeToDependedRector.php
samsonasik: instead, use ReturnsAnalyzer:

```php
        $returns = $this->betterNodeFinder->findReturnsScoped($functionLike);
        if (! $this->returnAnalyzer->hasOnlyReturnWithExpr($functionLike, $returns)) {
            return null;
        }
```

TomasVotruba: Good point :+1:

TomasVotruba:
Rector is failing because of different config locations. I'll add dummy set list to have them in relative paths

samsonasik:
The config is fine, just 'rector-src/vendor` is not exists in `rector-phpunit`, use `is_dir(__DIR__ . '/../../vendor)` check should work, I will create PR for it.

TomasVotruba:
@samsonasik No need to check magic paths, better use this way https://github.com/rectorphp/rector-symfony/pull/837

rules/CodeQuality/Rector/Class_/AddReturnTypeToDependedRector.php
```diff
@@ -0,0 +1,150 @@
...
+                continue;
+            }
+
+            if (! $this->testsNodeAnalyzer->isTestClassMethod($classMethod)) {
+                continue;
+            }
+
+            if (! $this->haveAllReturnsExpr($classMethod)) {
+                continue;
+            }
```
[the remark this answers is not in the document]
samsonasik: see https://github.com/rectorphp/rector-phpunit/pull/529

PR #529, opened 2025-09-01
Remove unused check return with expr as already covered on ReturnAnalyzer

samsonasik:
Ref https://github.com/rectorphp/rector-phpunit/pull/528#pullrequestreview-3174008146


TomasVotruba:
:+1:

PR #604, opened 2025-12-17
Failling test: MergeWithCallableAndWillReturnRector forgest to introduce closure `use`

staabm:
rector currently transforms the source into a syntax error 

grafik

because it forgets to introduce a `use ($initialStatements)` for the variable involved


staabm:
I don't plan to work on it, just reporting the bug

PR #666, opened 2026-04-01
[PHPUnit 9] Rename assertRegExp/assertNotRegExp to assertMatchesRegularExpression/assertDoesNotMatchRegularExpression on phpunit 90 set

### LINKED ISSUE #9714: Incorrect behavior of `AssertRegExpRector`
# Bug Report

| Subject        | Details              |
| :------------- | :--------------------|
| Rector version |   2.3.9   |
| Installed as   | composer dependency  |

See https://getrector.com/demo/43e1ad9c-0e8e-4d26-bc1e-e0a413e36cf0

```php
<?php

declare(strict_types=1);

namespace App\Tests\Integration\Video\Recreation;

use PHPUnit\Framework\Attributes\DataProvider;
use PHPUnit\Framework\Attributes\Test;

final class EpisodeRecreatorTest extends TestCase
{
    #[Test]
    public function recreate(): void
    {
        $function = fn (): string => 'something';
        
        self::assertSame(1, preg_match('/^.+__o__[0-9A-Fa-f]+__s\d+$/', $function()));
    }
}
```

### Responsible rules

* `AssertRegExpRector`

## Expected Behavior

`::assertRegExp` was [deprecated in v9](https://github.com/sebastianbergmann/phpunit/issues/4086) and removed in v10. 

It should use [`::assertMatchesRegularExpression()` instead](https://docs.phpunit.de/en/10.5/assertions.html#assertmatchesregularexpression).

samsonasik:
ok, on phpunit v8, `assertMatchesRegularExpression` and `assertDoesNotMatchRegularExpression` doesn't exists yet, so needs to apply on phpunit 9 set.

Fixes https://github.com/rectorphp/rector/issues/9714

### DIFF
--- a/config/sets/phpunit90.php
+++ b/config/sets/phpunit90.php
@@ -25,5 +25,20 @@ return static function (RectorConfig $rectorConfig): void {
             'expectExceptionMessageRegExp',
             'expectExceptionMessageMatches'
         ),
+
+
+        // @see https://github.com/sebastianbergmann/phpunit/issues/4086
+        new MethodCallRename(
+            'PHPUnit\Framework\TestCase',
+            'assertRegExp',
+            'assertMatchesRegularExpression'
+        ),
+
+        // @see https://github.com/sebastianbergmann/phpunit/issues/4089
+        new MethodCallRename(
+            'PHPUnit\Framework\TestCase',
+            'assertNotRegExp',
+            'assertDoesNotMatchRegularExpression'
+        ),
     ]);
 };


samsonasik:
Fixed 🎉 /cc @mvhirsch this should work with use phpunit 9 set so the `assertRegExp` will be renamed on version up to phpunit 9.

samsonasik:
@TomasVotruba ready 👍

TomasVotruba:
:+1:

mvhirsch:
> Fixed 🎉 /cc @mvhirsch this should work with use phpunit 9 set so the `assertRegExp` will be renamed on version up to phpunit 9.

@samsonasik I'm using PHPUnit v10 that's why it failed in first place (https://github.com/rectorphp/rector/issues/9714) :blush: 

Thank you!

mvhirsch:
@samsonasik but I wonder how this is going to fix `AssertRegExpRector` when running standalone as showcased in my reproducer https://github.com/rectorphp/rector/issues/9714?

I'm not using LevelSet updates, I'm just using this in my `rector.dist.php`:

```php
->withPreparedSets(deadCode: true, codeQuality: true, typeDeclarations: true, privatization: true, earlyReturn: true, phpunitCodeQuality: true, doctrineCodeQuality: true, symfonyCodeQuality: true)
```

https://github.com/rectorphp/rector-phpunit/blob/034f8a13b4b195b3479fcf30e3cbf6b628c6c0c8/rules/CodeQuality/Rector/MethodCall/AssertRegExpRector.php#L166

samsonasik:
You should can use `->withComposerBased(phpunit: true)` or use `->withSets([\Rector\PHPUnit\Set\PHPUnitSetList::PHPUNIT_90])`

The target method only exists in phpunit 9+ so use the composer based or use the setlist to update the name.

mvhirsch:
> The target method only exists in phpunit 9+

The target method is `assertMatchesRegularExpression`, but `AssertRegExpRector` hard-codes `assertRegExp` which was removed in v10. 

So if I understand you correctly, by using `->withPreparedSets(phpunitCodeQuality: true)` I'm implicitely forced to use `->withComposerBased(phpunit: true)` (or `->withSets(/* .. */)`) in order to tell Rector to use a non-removed method name of PHPUnit since years? I remember a blog post of @TomasVotruba using sets all the time is just a waste of time: running Set upgrades just once, while running quality of life fixes all the time. Maybe that knowledge is outdated? :thinking: 

I'm not familiar with internals of Rector. Is it common to use the _old_/deprecated method names in Rectors and rector them with level-sets afterwards on second iteration?

samsonasik:
I think the `AssertRegExpRector` can be moved to phpunit 9 rule set, and refactor to change regex function assert to rename to `assertMatchesRegularExpression`, @mvhirsch could you create PR for it? Thank you.

mvhirsch:
> @mvhirsch could you create PR for it? 

No, sorry.

samsonasik:
Ok, I created PR for it :)

- https://github.com/rectorphp/rector-phpunit/pull/667

PR #668, opened 2026-04-03
[CodeQuality] Rule to enforce test method code style

Xammie:
This adds two rules to convert the name of test methods to either `CamelCase` or `snake_case`.

```php
// When using
use Rector\PHPUnit\CodeQuality\Rector\ClassMethod\PreferTestsWithSnakeCaseRector
$rectorConfig->rule(PreferTestsWithSnakeCaseRector::class);

// From
public function testItWillCreateNewProduct() {}
// To
public function test_it_will_create_new_product() {}
```

### DIFF
--- a/rules/CodeQuality/Rector/Class_/PreferTestsWithCamelCaseRector.php
+++ b/rules/CodeQuality/Rector/Class_/PreferTestsWithCamelCaseRector.php
@@ -2,17 +2,17 @@
 
 declare(strict_types=1);
 
-namespace Rector\PHPUnit\CodeQuality\Rector\ClassMethod;
+namespace Rector\PHPUnit\CodeQuality\Rector\Class_;
 
 use PhpParser\Node;
-use PhpParser\Node\Stmt\ClassMethod;
+use PhpParser\Node\Stmt\Class_;
 use Rector\PHPUnit\NodeAnalyzer\TestsNodeAnalyzer;
 use Rector\Rector\AbstractRector;
 use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
 use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
 
 /**
- * @see \Rector\PHPUnit\Tests\CodeQuality\Rector\ClassMethod\PreferTestsWithCamelCaseRector\PreferTestsWithCamelCaseRectorTest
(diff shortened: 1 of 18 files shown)


Xammie:
@TomasVotruba this is ready to review :)

rules/CodeQuality/Rector/Class_/PreferTestsWithCamelCaseRector.php
```diff
@@ -0,0 +1,93 @@
...
+    public function refactor(Node $node): ?Node
+    {
+        if (! $this->testsNodeAnalyzer->isInTestClass($node)) {
+            return null;
+        }
+
+        if (! $this->testsNodeAnalyzer->isTestClassMethod($node)) {
+            return null;
+        }
+
```
samsonasik, line 71: here ensure name is exists first before changing it, you can use `Class_` instead of `ClassMethod`, and verify from `$class->getMethod()` or `classReflection->hasMethod()`

samsonasik:
for additional note: I think this should be your own custom rule for your specific use case :)

Xammie:
@samsonasik thank you for the review. 

Regarding your additional note, do you mean this should not be part of rector php and instead be part of a separate package?

samsonasik:
yeah, imo, this should be not part of rector repo, you can create your own custom rector rule for it as this is by preference

Xammie:
The reason I thought it would fit well here is because this is similar to [`PreferPHPUnitSelfCallRector`](https://getrector.com/rule-detail/prefer-p-h-p-unit-self-call-rector) and [`PreferPHPUnitThisCallRector`](https://getrector.com/rule-detail/prefer-p-h-p-unit-this-call-rector) in a way, because that is mostly a preference in code style.

Ofcourse if you still don't think this should be part of `rector-phpunit` then I completely understand and I will put it into a new package.

TomasVotruba:
I'm looking into this. Seems valid to have the direction as well :+1:

rules/CodeQuality/Rector/Class_/PreferTestsWithCamelCaseRector.php
TomasVotruba: Addressed: both rules now operate on `Class_` node type and iterate `$class->getMethods()`. Before renaming, they check `$class->getMethod($newName)` and skip when a method with the target name already exists, to avoid name collisions. Added `skip_existing_name` fixtures covering this. Rules moved to the `Rector/Class_/` directory accordingly.

TomasVotruba:
Ready to ship :+1:  Thank you!

PR #669, opened 2026-04-29
[CodeQuality] Skip multiple uses variable as arg after defined on BareCreateMockAssignToDirectUseRector

samsonasik:
When used multiple times as argument value, it can already changed on caller, so this should be skipped.

```php
        $someMock = $this->createMock(SomeClass::class);

        // behaviour can already changed on first use
        $this->useMock($someMock);
        $this->useMockAgain($someMock);
```

This PR skip it, this example happen in our project ;)


samsonasik:
@TomasVotruba ready 👍

TomasVotruba:
LGTM :+1:

PR #670, opened 2026-04-29
[PHPUnit12] Skip used as next arg with MockObject Type on ExpressionCreateMockToCreateStubRector

samsonasik:
When used as `MockObject` on next arg with `MockObject` type, it should be skipped:

```php
    public function test()
    {
        $mock = $this->createMock(ClassWithDependency::class);
        $this->configureMock($mock);
    }

    private function configureMock(\PHPUnit\Framework\MockObject\MockObject $mock)
    {
        $mock->method('someMethod')->willReturn(1);
    }
```

we have this case on our project ;)


samsonasik:
@TomasVotruba ready 👍

TomasVotruba (approved):

TomasVotruba:
LGTM

PR #671, opened 2026-05-11
ReplaceTestAnnotationWithPrefixedFunctionRector does not match @test annotation if it is being part of string

Dukecz:
Fix for: https://github.com/rectorphp/rector/issues/9752

### DIFF
--- a/rules/CodeQuality/Rector/ClassMethod/ReplaceTestAnnotationWithPrefixedFunctionRector.php
+++ b/rules/CodeQuality/Rector/ClassMethod/ReplaceTestAnnotationWithPrefixedFunctionRector.php
@@ -12,9 +12,9 @@ use Rector\BetterPhpDocParser\PhpDocManipulator\PhpDocTagRemover;
 use Rector\Comments\NodeDocBlock\DocBlockUpdater;
 use Rector\PHPUnit\NodeAnalyzer\TestsNodeAnalyzer;
 use Rector\Rector\AbstractRector;
+use Rector\Util\NewLineSplitter;
 use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
 use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
-use function explode;
 use function in_array;
 
 /**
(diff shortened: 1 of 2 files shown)


rules/CodeQuality/Rector/ClassMethod/ReplaceTestAnnotationWithPrefixedFunctionRector.php
```diff
@@ -84,7 +86,14 @@ public function refactor(Node $node): ?Node
             return null;
         }
 
-        if (! str_contains($docComment->getText(), '@test')) {
+        $hasAnnotation = false;
+        foreach(explode(PHP_EOL, $docComment->getText()) as $row) {
```
samsonasik: This may not compatible when different new line opened in different operating system. You can use 

```php
\Rector\Util\NewLineSplitter::split($docComment->getText())
```

instead.

Dukecz: I see. Fixed in https://github.com/rectorphp/rector-phpunit/pull/671/changes/c0ef6a53320f90706ae68bab19793fcebeb7de22

rules/CodeQuality/Rector/ClassMethod/ReplaceTestAnnotationWithPrefixedFunctionRector.php
```diff
@@ -84,7 +86,14 @@ public function refactor(Node $node): ?Node
             return null;
         }
 
-        if (! str_contains($docComment->getText(), '@test')) {
+        $hasAnnotation = false;
+        foreach(NewLineSplitter::split($docComment->getText()) as $row) {
+            if (in_array(trim($row), ['*@test', '* @test'])) {
+                $hasAnnotation = true;
```
samsonasik, line 92: break when mark as `true` to stop unnecessary next iteration.

```suggestion
                $hasAnnotation = true;
                break;
```

samsonasik (approved):

TomasVotruba:
Thank you ☝️

PR #672, opened 2026-05-11
[CodeQuality] Handle on static closure on WithCallbackIdenticalToStandaloneAssertsRector

### LINKED ISSUE #9754: Incorrect behavior of WithCallbackIdenticalToStandaloneAssertsRector with static anonymous function
# Bug Report

| Subject        | Details              |
| :------------- | :--------------------|
| Rector version | 2.4.2      |
| Installed as   | composer dependency  |

See https://getrector.com/demo/d8eb040a-bade-476f-80db-31a111b7720c

```php
<?php

final class SomeTest extends TestCase
{
    public function test()
    {
        $this->createMock('SomeClass')
            ->expects($this->once())
            ->method('someMethod')
            ->with($this->callback(static function (array $args): bool {
               return count($args) === 2 && $args[0] === 'correct';
            }));
    }
}
```

### Responsible rules

* `WithCallbackIdenticalToStandaloneAssertsRector`

## Expected Behavior

If the anonymouse function is static, it should also remove the static keyword, as otherwise the changed code containing the $this variable will break.

Just a minor thing. I think removing the static keyword should always be fine in this case, but maybe some deeper analysis is needed? Otherwise the rule should maybe ignore static functions.

samsonasik:
Fixes https://github.com/rectorphp/rector/issues/9754

### DIFF
--- a/rules-tests/CodeQuality/Rector/MethodCall/WithCallbackIdenticalToStandaloneAssertsRector/Fixture/on_static_closure.php.inc
+++ b/rules-tests/CodeQuality/Rector/MethodCall/WithCallbackIdenticalToStandaloneAssertsRector/Fixture/on_static_closure.php.inc
@@ -1,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\PHPUnit\Tests\CodeQuality\Rector\MethodCall\WithCallbackIdenticalToStandaloneAssertsRector\Fixture;
+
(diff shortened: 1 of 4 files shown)


samsonasik:
Fixed 🎉 /cc @LorisZ

samsonasik:
@TomasVotruba ready 👍

TomasVotruba (approved):

TomasVotruba:
Thanks

PR #674, opened 2026-05-21
fix AssertIsTypeMethodCallRector crash on enum case argument

### LINKED ISSUE #9765: AssertIsTypeMethodCallRector crashes on 'isType' method call on any other object
# Bug Report

| Subject        | Details                                            |
| :------------- | :------------------------------------------------- |
| Rector version | 2.4.2 |
| PHP version | 8.5.2 |

`$someObject->isType()` inside a subclass of `PHPUnit\Framework\TestCase` crashes the `AssertIsTypeMethodCallRector` when an enum case is passed in.

Error:
```
[ERROR] Could not process
         "thefile.php" file, due to:
         "System error: "Cannot access offset of type App\Enums\SomeEnum in isset or empty"
         Run Rector with "--debug" option and post the report here: https://github.com/rectorphp/rector/issues/new". On line: 92
```

The violating file:
```php
<?php

namespace Tests;

use App\Enums\SomeEnum;
use stdClass;

final class MyTest extends TestCase
{
    public function test_ing(): void
    {
        (new stdClass)->isType(SomeEnum::ONE);
    }
}
```

With enum:
```php
<?php

namespace App\Enums;

enum SomeEnum: string
{
    case ONE = 'one';
}
```

Rector.php:
```php
<?php

use Rector\Config\RectorConfig;
use Rector\TypeDeclaration\Rector\Property\TypedPropertyFromAssignsRector;

return RectorConfig::configure()
    ->withRules([
        \Rector\PHPUnit\PHPUnit120\Rector\Class_\AssertIsTypeMethodCallRector::class
    ]);
```

> https://getrector.com/demo/ can't reproduce it since:
> 1. The Enum should be in another namespace in another file (not sure why it doesn't crash when the enum is in the same file)
> 2. The demo can't check if a class is a phpunit testcase

## Expected Behaviour

Rector shouldn't crash and shouldn't have anything to do with this specific case since the `isType` method is not called on `$this`.

TomasVotruba:
## Summary

- Fixes a fatal crash in `AssertIsTypeMethodCallRector` when `isType()` is called with an enum case argument inside a `TestCase` subclass
- Adds a `is_string()` guard before using the resolved arg value as an array key in `isset()`
- Adds a skip fixture to cover the non-string arg scenario

## Root cause

`ValueResolver::getValue()` returns the enum case object itself for enum arguments. Using it as an array key in `isset(self::IS_TYPE_VALUE_TO_METHOD[$argValue])` caused:

```
System error: "Cannot access offset of type App\Enums\SomeEnum in isset or empty"
```

## Fix

Added an `is_string($argValue)` guard immediately after resolving the argument value. Non-string values can never match any key in `IS_TYPE_VALUE_TO_METHOD`, so returning `null` early is both correct and safe.

Fixes https://github.com/rectorphp/rector/issues/9765

---
_Generated by [Claude Code](https://claude.ai/code/session_013NBfVf3ZKpBieeTf4xxAji)_

### DIFF
--- a/rules-tests/PHPUnit120/Rector/Class_/AssertIsTypeMethodCallRector/Fixture/skip_non_string_arg.php.inc
+++ b/rules-tests/PHPUnit120/Rector/Class_/AssertIsTypeMethodCallRector/Fixture/skip_non_string_arg.php.inc
@@ -1,0 +1,20 @@
+<?php
+
+// skip: non-string argument (enum case) should not crash
+
+namespace Rector\PHPUnit\Tests\PHPUnit120\Rector\Class_\AssertIsTypeMethodCallRector\Fixture;
+
+use PHPUnit\Framework\TestCase;
+
+enum SkipNonStringArgEnum
+{
+    case ONE;
+}
+
+final class SkipNonStringArg extends TestCase
+{
+    public function testMethod(): void
+    {
+        $this->assertThat([], $this->isType(SkipNonStringArgEnum::ONE));
+    }
+}
--- a/rules/PHPUnit120/Rector/Class_/AssertIsTypeMethodCallRector.php
+++ b/rules/PHPUnit120/Rector/Class_/AssertIsTypeMethodCallRector.php
@@ -114,6 +114,10 @@ final class AssertIsTypeMethodCallRector extends AbstractRector
         }
 
         $argValue = $this->valueResolver->getValue($arg);
+        if (! is_string($argValue)) {
+            return null;
+        }
+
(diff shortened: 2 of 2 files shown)


TomasVotruba:
Looks good 👌

Levivb:
not really though ;)

although the actual crash is fixed, the underlying issue is that `isType` is processed regardless of the classtype the method is called on.

```diff
- (new stdClass)->isType('array');
+ (new stdClass)->isArray();
```

TomasVotruba:
2 bugs in one :)

PR #683, opened 2026-06-26
[CodeQuality] Fix MergeWithCallableAndWillReturnRector to add closure use() for variables nested in willReturn expr

TomasVotruba:
Previously only a bare Variable willReturn arg got a closure use(). A variable nested inside an array (e.g. willReturn([$initialStatements, []])) was moved into the closure without a use(), producing a syntax error.

Now all variables referenced in the returned expression are collected (skipping closure params and $this) and added as closure uses.

Fixes rectorphp/rector-phpunit#604


staabm:
thank you!

PR #687, opened 2026-06-29
[PHPUnit60] Fix false positive in AddDoesNotPerformAssertionToNonAssertingTestRector

TomasVotruba:
## What

`AssertCallAnalyzer::containsAssertCall()` incremented `$classMethodNestingLevel` on every call but **never decremented it**. So instead of tracking real recursion depth, the counter accumulated across the whole depth-first walk of a single test method.

## Why it caused false positives

When a non-asserting call (e.g. a production static call) was visited *before* an assert-performing helper, it burned through the level-5 budget. The later helper recursion then hit `level > MAX_NESTED_METHOD_CALL_LEVEL` and short-circuited to `false`, so a test that *does* assert via a helper was wrongly tagged `@doesNotPerformAssertions`.

Order-dependent: any test where a non-asserting call precedes the assert helper in source order.

Found while investigating why `webonyx/graphql-php` skips this rule as "False-positive". Running the rule there flagged 7+ tests that assert through helpers (`$this->runTestCase()`, `$this->expectGraphQLValue()`, `$this->checkHandlesNonNullableLists()`, ...). With this fix only genuine `markTestSkipped()` tests remain.

## Fix

Wrap the body in `try { ... } finally { --$this->classMethodNestingLevel; }` so the depth is restored on unwind — sibling calls in the same DFS keep their full budget.

## Test

Added keep fixture `keep_deep_calls_before_assert_helper.php.inc`: a deep non-asserting chain visited before an assert helper. Fails on the old code (annotation added), passes with the fix.

### DIFF
--- a/rules-tests/PHPUnit60/Rector/ClassMethod/AddDoesNotPerformAssertionToNonAssertingTestRector/Fixture/keep_deep_calls_before_assert_helper.php.inc
+++ b/rules-tests/PHPUnit60/Rector/ClassMethod/AddDoesNotPerformAssertionToNonAssertingTestRector/Fixture/keep_deep_calls_before_assert_helper.php.inc
@@ -1,0 +1,47 @@
+<?php
+
+namespace Rector\PHPUnit\Tests\PHPUnit60\Rector\ClassMethod\AddDoesNotPerformAssertionToNonAssertingTestRector\Fixture;
+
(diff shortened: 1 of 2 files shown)


TomasVotruba:
This might be a perf bottleneck, but lets give it a go. We can adjust nested level to lower value if needed.

PR #705, opened 2026-07-07
[AnnotationsToAttributes] Avoid partial apply on CoversAnnotationWithValueToAttributeRector

samsonasik:
When apply partial migration, this rule may cause error:

```
"@covers ::render" is invalid
```

this PR make early check that all possible migration class are there first so no more partial apply due to PHPUnit minor version introduce new attribute classes.

### DIFF
--- a/rules/AnnotationsToAttributes/Rector/Class_/CoversAnnotationWithValueToAttributeRector.php
+++ b/rules/AnnotationsToAttributes/Rector/Class_/CoversAnnotationWithValueToAttributeRector.php
@@ -108,7 +108,13 @@ CODE_SAMPLE
             return null;
         }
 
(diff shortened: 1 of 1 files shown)


samsonasik:
@TomasVotruba ready 👍

TomasVotruba:
Can you add test fixture as well?

samsonasik:
not sure, as the issue happen on phpunit 10, and this repo already uses phpunit 12, so it always work ok on phpunit 12 as already all there.

The example use is on spiral framework:

https://github.com/spiral/framework/blob/af6447ddf37d7236b9851d009888bc716ada7f09/src/SendIt/tests/ViewRendererTest.php#L27

TomasVotruba:
I see 👌

PR #726, opened 2026-07-13
[CodeQuality] Skip new object instances in NarrowIdenticalWithConsecutiveRector

TomasVotruba:
Skip narrowing `withConsecutive()` / `willReturnOnConsecutiveCalls()` when any consecutive arg builds a `new` object.

Each `new` call creates a fresh instance, possibly with a property/constructor dependency, so collapsing them to a single `with()` / `willReturn()` changes intent.

```diff
 $someServiceMock->expects($this->exactly(3))
     ->method('prepare')
-    ->withConsecutive(
-        [new \stdClass()],
-        [new \stdClass()],
-        [new \stdClass()],
-    );
+    ->withConsecutive(
+        [new \stdClass()],
+        [new \stdClass()],
+        [new \stdClass()],
+    ); // now skipped, left as-is
```

Plain scalar repeats still narrow as before:

```diff
 $someServiceMock->expects($this->exactly(3))
-    ->method('prepare')
-    ->withConsecutive([1], [1], [1]);
+    ->method('prepare')->with([1]);
```


TomasVotruba:
To avoid side-effects, that we found in one project when new statefull class was created.

PR #770, opened 2026-08-13
[CodeQuality] Narrow assertTrue(is_<type>()) to assertIs<Type>()

Soean:
`AssertTrueFalseToSpecificMethodRector` already narrows `assertTrue(is_readable(...))`, `assertTrue(is_null(...))` and friends, but the `is_<type>()` family was missing — so `$this->assertTrue(is_callable($value))` stayed as a generic boolean assertion instead of becoming `$this->assertIsCallable($value)`.
The generic form loses the failure message: `assertTrue()` can only report "false is not true", while `assertIsCallable()` reports the actual value and the type it was expected to have.

## Added mappings
| function | assertTrue / assertNotFalse | assertFalse / assertNotTrue |
| --- | --- | --- |
| `is_array` | `assertIsArray` | `assertIsNotArray` |
| `is_bool` | `assertIsBool` | `assertIsNotBool` |
| `is_callable` | `assertIsCallable` | `assertIsNotCallable` |
| `is_float` | `assertIsFloat` | `assertIsNotFloat` |
| `is_int` | `assertIsInt` | `assertIsNotInt` |
| `is_iterable` | `assertIsIterable` | `assertIsNotIterable` |
| `is_numeric` | `assertIsNumeric` | `assertIsNotNumeric` |
| `is_object` | `assertIsObject` | `assertIsNotObject` |
| `is_scalar` | `assertIsScalar` | `assertIsNotScalar` |
| `is_string` | `assertIsString` | `assertIsNotString` |

Each of the ten was checked against `IsType::matches()` to confirm the constraint runs the very same function, so the assertions are equivalent.

`is_callable()` is skipped when it carries more than one argument — its `$syntax_only` and `$callable_name` parameters have no counterpart in `assertIsCallable()` and would otherwise be moved up into the `$message` slot.

## Also in here: named and spread args

Args are moved up by position, so a named arg only survives when the assert method happens to use the same parameter name. That holds for `is_readable(filename:)`, `is_writable(filename:)`, `file_exists(filename:)`, `array_key_exists(key:, array:)` and `str_contains(haystack:, needle:)`. It does not hold for any of the new entries, as PHP names the parameter `$value` while PHPUnit names it `$actual`:

    $this->assertTrue(is_string(value: $value));
    // became assertIsString(value: $value) -> Unknown named parameter $value

The bail-out is blanket rather than per entry, which also covers `is_dir` (`$filename` vs `$directory`), `is_null`, `is_nan`, `is_infinite` and `is_a` — those carry the same mismatch today. The cost is that the handful of pairs whose names do line up get skipped too, which seemed a better trade than tracking a parameter name per mapped function. Spread args are skipped for the same reason, as their count is unknown.

It sits in its own `hasUnmovableArgs()` method, so it is easy to pull out if you would rather keep this PR to the new mappings.

### DIFF
--- a/rules/CodeQuality/Rector/MethodCall/AssertTrueFalseToSpecificMethodRector.php
+++ b/rules/CodeQuality/Rector/MethodCall/AssertTrueFalseToSpecificMethodRector.php
@@ -171,8 +171,11 @@ final class AssertTrueFalseToSpecificMethodRector extends AbstractRector
     }
 
     /**
-     * The assert methods use different parameter names than the functions they replace,
-     * so named args cannot be moved up. The same goes for spread args, as their count is unknown.
+     * Args are moved up by position, so a named arg only survives when the assert method happens to
+     * use the same parameter name: is_readable(filename:) fits assertIsReadable(), while
+     * is_string(value:) does not fit assertIsString(), whose parameter is $actual. Bail out instead
+     * of tracking a parameter name per mapped function. Spread args cannot be moved up either,
+     * as their count is unknown.
      */
     private function hasUnmovableArgs(FuncCall $funcCall): bool
     {
--- a/rules/CodeQuality/Rector/MethodCall/AssertTrueFalseToSpecificMethodRector.php
+++ b/rules/CodeQuality/Rector/MethodCall/AssertTrueFalseToSpecificMethodRector.php
@@ -179,6 +179,10 @@ final class AssertTrueFalseToSpecificMethodRector extends AbstractRector
      */
     private function hasUnmovableArgs(FuncCall $funcCall): bool
     {
+        if ($funcCall->isFirstClassCallable()) {
+            return true;
+        }
+
         foreach ($funcCall->getArgs() as $arg) {
             if ($arg->name instanceof Identifier) {
                 return true;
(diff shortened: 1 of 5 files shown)


rules/CodeQuality/Rector/MethodCall/AssertTrueFalseToSpecificMethodRector.php
```diff
@@ -148,6 +170,28 @@ private function resolveFirstArgument(FuncCall|Empty_ $firstArgumentValue): ?str
...
 
+    /**
+     * Args are moved up by position, so a named arg only survives when the assert method happens to
+     * use the same parameter name: is_readable(filename:) fits assertIsReadable(), while
+     * is_string(value:) does not fit assertIsString(), whose parameter is $actual. Bail out instead
+     * of tracking a parameter name per mapped function. Spread args cannot be moved up either,
+     * as their count is unknown.
+     */
+    private function hasUnmovableArgs(FuncCall $funcCall): bool
+    {
```
samsonasik, line 181: here check `$funcCall->isFirstClassCallable()` early to avoid warning on first class callable

```suggestion
    {
        if ($funcCall->isFirstClassCallable()) {
            return true;
        }
```

TomasVotruba: Done, added the `isFirstClassCallable()` guard early in `hasUnmovableArgs()`, plus a `is_string(...)` skip fixture.

TomasVotruba:
Thank you :clap:

PR #771, opened 2026-08-20
[CodeQuality] Fix MatchAssertSameExpectedTypeRector rewriting a passing assert into a failing one

MizouziE:
`MatchAssertSameExpectedTypeRector` can rewrite a passing `assertSame()` into a failing one.

## Mechanism

The rule picks the expected literal's new type from `NodeTypeResolver::getNativeType()`, which returns PHPStan's **scope-narrowed** type at that program point. A narrowed type is sound to report on, but not to rewrite a literal against: it describes the expression where the narrowing was established, while the rewrite lands elsewhere.

Given `Criteria::getType(): int|string|null`:

```php
$this->assertSame('Any', $criteria->getType());  // phpstan-phpunit narrows getType() to string

$criteria->setTypeId(100);                       // mutates the receiver
$this->assertSame(100, $criteria->getType());    // rewritten to '100' and fails
```


TomasVotruba:
Looks good, thank you :+1:

PR #773, opened 2026-08-25
[CodeQuality] Fix AddInstanceofAssertForNullableArgumentRector duplicate assertInstanceOf inside traits

### LINKED ISSUE #9863: [PHPUnit] AddInstanceofAssertForNullableArgumentRector adds a duplicate assertInstanceOf on every run inside traits
# Bug Report

| Subject | Details |
| :------ | :------ |
| Rector version | `rector/rector` dev-main (`61a62fa`), bundled `rector/rector-phpunit` dev-main (`0f24ff4`) |
| PHP version | 8.5.9 |
| Rule | `AddInstanceofAssertForNullableArgumentRector` |

## Summary

Inside a **test trait**, `AddInstanceofAssertForNullableArgumentRector` is not idempotent. Every `rector process` run appends one more identical `$this->assertInstanceOf(...)` line, forever.

In our codebase this went unnoticed across a few Rector runs and a single method ended up with nine copies of the same assertion:

```php
$item = $conformingItem->fresh();
$this->assertInstanceOf(Item::class, $item);
$this->assertInstanceOf(Item::class, $item);
$this->assertInstanceOf(Item::class, $item);
$this->assertInstanceOf(Item::class, $item);
$this->assertInstanceOf(Item::class, $item);
$this->assertInstanceOf(Item::class, $item);
$this->assertInstanceOf(Item::class, $item);
$this->assertInstanceOf(Item::class, $item);
$this->assertInstanceOf(Item::class, $item);
```

The equivalent code in a class extending `TestCase` is correctly idempotent — it only reproduces in a trait.

## Minimal reproducer

`rector.php`:

```php
<?php

declare(strict_types=1);

use Rector\Config\RectorConfig;
use Rector\PHPUnit\CodeQuality\Rector\ClassMethod\AddInstanceofAssertForNullableArgumentRector;

return RectorConfig::configure()
    ->withPaths([__DIR__ . '/src'])
    ->withRules([AddInstanceofAssertForNullableArgumentRector::class]);
```

`src/SomeAssertionsTrait.php`:

```php
<?php

namespace App\Tests;

final class SomeClass
{
}

final class SomeFactory
{
    public function create(): ?SomeClass
    {
        return null;
    }

    public function process(SomeClass $someClass): void
    {
    }
}

trait SomeAssertionsTrait
{
    public function assertSomething(SomeFactory $someFactory): void
    {
        $someObject = $someFactory->create();

        $someFactory->process($someObject);
    }
}
```

Then run `vendor/bin/rector process --clear-cache` three times.

Note the consuming call must be on something other than `$this` (here `$someFactory->process()`). When the call is `$this->process()`, the rule does not fire at all in a trait, because the parameter types cannot be resolved through `$this`.

## Expected behaviour

Run 1 adds the assertion; runs 2 and 3 report no change.

## Actual behaviour

Every run adds another copy:

```diff
=== RUN 1 === changed: 1
         $someObject = $someFactory->create();
+        $this->assertInstanceOf(\App\Tests\SomeClass::class, $someObject);

         $someFactory->process($someObject);

=== RUN 2 === changed: 1
         $someObject = $someFactory->create();
         $this->assertInstanceOf(\App\Tests\SomeClass::class, $someObject);
+        $this->assertInstanceOf(\App\Tests\SomeClass::class, $someObject);

         $someFactory->process($someObject);

=== RUN 3 === changed: 1
         $someObject = $someFactory->create();
         $this->assertInstanceOf(\App\Tests\SomeClass::class, $someObject);
         $this->assertInstanceOf(\App\Tests\SomeClass::class, $someObject);
+        $this->assertInstanceOf(\App\Tests\SomeClass::class, $someObject);

         $someFactory->process($someObject);
```

## Analysis

The rule has no check for an assertion it already added. Its only guard against duplicates is within a single run — `VariableNameToTypeCollection::remove()` in `AddInstanceofAssertForNullableArgumentRector::refactor()`:

```php
// from now on, the variable is not nullable, remove to avoid double asserts
$variableNameToTypeCollection->remove($matchedNullableVariableNameToType);
```

Across runs, the rule relies on the inserted `$this->assertInstanceOf()` narrowing the variable's type so that `SimpleTypeAnalyzer::isNullableType()` no longer matches. In a class extending `TestCase` that holds, and the rule is stable. In a trait it does not: `$this` is not a `PHPUnit\Framework\Assert`, so the assertion does not narrow the type, the variable stays nullable on the next run, and a fresh assertion is appended.

`isInTestClass()` grew explicit trait support in #751 (`Handle test traits in DataProviderAnnotationToAttributeRector`), which routes traits through `isInTestTrait()` and matches on a `Test`/`Tests` namespace segment. That made this rule apply inside traits, where its type-narrowing assumption does not hold.

I verified the split by running the same method body in three shapes:

| Shape | Behaviour |
| :---- | :-------- |
| `final class ... extends TestCase` | adds once, then stable |
| `trait`, nullable from `?SomeClass` | one duplicate per run |
| `trait`, nullable from `?static` | one duplicate per run |

So the trait context is the trigger, not the nullable return type.

## Suggested fix

Do not depend on type narrowing for idempotency. Before splicing, check whether the statement list already contains an equivalent `assertInstanceOf(<Type>::class, $variable)` for that variable and skip if so. That fixes the trait case and also hardens the class case against any setup where the PHPUnit type-specifying extension is unavailable.

Alternatively, skip traits in this rule, since neither the narrowing nor the `$this->` parameter-type resolution works reliably there.

TomasVotruba:
Fixes rectorphp/rector#9863

## Problem

Inside a **test trait**, `AddInstanceofAssertForNullableArgumentRector` is not idempotent. Every `rector process` run appends one more identical `$this->assertInstanceOf(...)` line, forever.

The rule's only cross-run guard was implicit: the inserted `$this->assertInstanceOf()` narrows the variable's type, so `SimpleTypeAnalyzer::isNullableType()` stops matching on the next run. In a class extending `TestCase` that holds. In a trait `$this` is not a `PHPUnit\Framework\Assert`, so the assert never narrows, the variable stays nullable, and a fresh assert is appended each run.

```diff
 $someObject = $someFactory->create();
 $this->assertInstanceOf(SomeClass::class, $someObject);
+$this->assertInstanceOf(SomeClass::class, $someObject);   // added again every run
 $someFactory->process($someObject);
```

## Fix

Stop depending on type narrowing for idempotency. While scanning statements, detect a pre-existing `assertInstanceOf(<Type>::class, $variable)` and drop that variable from the nullable collection, so no second assert is spliced in. Works in both classes and traits, and hardens the class case when the PHPUnit type-specifying extension is unavailable.

## Tests

- `skip_already_asserted_in_trait.php.inc` — trait already carrying the assert stays unchanged (fails before the fix: a duplicate assert is added).
- `nullable_arg_in_trait.php.inc` — the rule still fires once in a trait.

### DIFF
--- a/rules-tests/CodeQuality/Rector/ClassMethod/AddInstanceofAssertForNullableArgumentRector/Fixture/nullable_arg_in_trait.php.inc
+++ b/rules-tests/CodeQuality/Rector/ClassMethod/AddInstanceofAssertForNullableArgumentRector/Fixture/nullable_arg_in_trait.php.inc
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\PHPUnit\Tests\CodeQuality\Rector\ClassMethod\AddInstanceofAssertForNullableArgumentRector\Fixture;
+
+use Rector\PHPUnit\Tests\CodeQuality\Rector\ClassMethod\AddInstanceofAssertForNullableArgumentRector\Source\SomeFactory;
+
+trait NullableArgInTrait
+{
+    public function assertSomething(SomeFactory $someFactory): void
+    {
+        $someObject = $someFactory->create();
+
+        $someFactory->process($someObject);
+    }
+}
+
+?>
+-----
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\PHPUnit\Tests\CodeQuality\Rector\ClassMethod\AddInstanceofAssertForNullableArgumentRector\Fixture;
+
+use Rector\PHPUnit\Tests\CodeQuality\Rector\ClassMethod\AddInstanceofAssertForNullableArgumentRector\Source\SomeFactory;
+
+trait NullableArgInTrait
+{
+    public function assertSomething(SomeFactory $someFactory): void
+    {
+        $someObject = $someFactory->create();
+        $this->assertInstanceOf(\Rector\PHPUnit\Tests\CodeQuality\Rector\ClassMethod\AddInstanceofAssertForNullableArgumentRector\Source\SomeClassUsedInTests::class, $someObject);
+
+        $someFactory->process($someObject);
+    }
+}
+
+?>
--- a/rules-tests/CodeQuality/Rector/ClassMethod/AddInstanceofAssertForNullableArgumentRector/Fixture/skip_already_asserted_in_trait.php.inc
+++ b/rules-tests/CodeQuality/Rector/ClassMethod/AddInstanceofAssertForNullableArgumentRector/Fixture/skip_already_asserted_in_trait.php.inc
@@ -1,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\PHPUnit\Tests\CodeQuality\Rector\ClassMethod\AddInstanceofAssertForNullableArgumentRector\Fixture;
+
+use Rector\PHPUnit\Tests\CodeQuality\Rector\ClassMethod\AddInstanceofAssertForNullableArgumentRector\Source\SomeFactory;
+
+trait SkipAlreadyAssertedInTrait
+{
+    public function assertSomething(SomeFactory $someFactory): void
+    {
+        $someObject = $someFactory->create();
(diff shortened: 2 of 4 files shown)


TomasVotruba:
Let's ship it :+1:

PR #779, opened 2026-09-01
 [CodeQuality] Skip existing imported or short @ see annotation in AddSeeTestAnnotationRector

### LINKED ISSUE #9875: AddSeeTestAnnotationRector adds a duplicate @see when the existing annotation is imported
# Bug Report

| Subject | Details |
| :--- | :--- |
| Rector version | v2.6.5 |

`AddSeeTestAnnotationRector` adds a second `@see` annotation when the first one is imported rather than fully qualified.

The rule writes `@see \App\Tests\FooTest`. php-cs-fixer's `fully_qualified_strict_types`, which PER-CS enables, then shortens that to `@see FooTest` and adds a `use`. On the next run the rule no longer recognizes its own annotation and appends another, so the count grows on every run.

`hasAlreadySeeAnnotation()` compares the raw docblock text against the fully qualified name, so the short form never matches:

https://github.com/rectorphp/rector-phpunit/blob/c4e571ee6d61aa0bf68346bab12b85e2df3aead2/rules/CodeQuality/Rector/Class_/AddSeeTestAnnotationRector.php#L164-L169

This was reported once before in rectorphp/rector#9795 and closed because the reporter's test classes turned out to be in the wrong namespace, so the reproduction never held up. The one below does — the test class is found, which is what makes the rule add the duplicate.

With `App\Tests\FooTest` present and `rectorPreset: true`:

```php
namespace App;

use App\Tests\FooTest;

/**
 * @see FooTest
 */
final class Foo {}
```

```diff
 /**
  * @see FooTest
+ * @see \App\Tests\FooTest
  */
```

## Expected Behaviour

Skip the class — the annotation is already there, just imported. Resolving the name against the file's `use` statements before comparing would do it.

samsonasik:
Fixes https://github.com/rectorphp/rector/issues/9875

### DIFF
--- a/rules-tests/CodeQuality/Rector/Class_/AddSeeTestAnnotationRector/Fixture/skip_existing_imported_see.php.inc
+++ b/rules-tests/CodeQuality/Rector/Class_/AddSeeTestAnnotationRector/Fixture/skip_existing_imported_see.php.inc
@@ -1,0 +1,12 @@
+<?php
+
+namespace Rector\PHPUnit\Tests\CodeQuality\Rector\Class_\AddSeeTestAnnotationRector\Fixture;
+
+use Rector\PHPUnit\Tests\CodeQuality\Rector\Class_\AddSeeTestAnnotationRector\Source\SkipExistingImportedSeeTest;
+
+/**
+ * @see SkipExistingImportedSeeTest
+ */
+class SkipExistingImportedSee
+{
+}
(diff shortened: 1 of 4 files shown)


samsonasik:
Fixed 🎉 /cc @Brainshaker95

samsonasik:
@TomasVotruba ready 👍

TomasVotruba:
Thanks

PR #783, opened 2026-09-17
Fix changing keys of data providers for tests with optional arguments

tunterreitmeier:
Hi, I noticed that the `NamedArgumentForDataProviderRector` currently tries to fix named arguments, even when every key matches an argument. This can break tests, when there are optional arguments involved.

So when running the rule on this example
```php
final class TestWithOptionalArguments extends TestCase
{
    public static function dataProvider(): Generator
    {
        yield [
            'keyA' => false,
            'keyC' => true,
        ];
    }

    #[DataProvider('dataProvider')]
    public function testFilter(bool $keyA, bool $keyB = false, bool $keyC = false): void
    {
    }
}
```

one would end up with this diff:

```diff
yield [
  'keyA' => false,
-  'keyC' => true,
+  'keyB' => true,
];
```

My proposed fix would be to skip if all the keys are already all named arguments, using `array_diff`.

Hope that helps. Thanks a lot!


TomasVotruba:
LGTM, Thank you :+1:

